Issue #4, opened 2018-01-29
Support 'checkbox' input type in the crawler


marco-c:
https://github.com/marco-c/autowebcompat/blob/e3f77c7b48e017a0ddfe0116e8748964d36fa74c/collect.py#L104

SkullTech:
This issue should be closed as it was solved by #8

PR #8, opened 2018-01-30
Support input type

sagarvijaygupta:
Support 'number' input type in the crawler
Support 'checkbox' input type in the crawler

Solves : #3 and #4

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -111,6 +111,10 @@ def do_something(driver, elem_id=None):
             elem.send_keys('prova@email.it')
         elif input_type == 'password':
             elem.send_keys('aMildlyComplexPasswordIn2017')
+        elif input_type == 'checkbox':
+            elem.click()
+        elif input_type == 'number':
+            elem.send_keys('3')
         else:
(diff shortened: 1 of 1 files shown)


sagarvijaygupta:
checkbox_unchecked
checkbox_checked
number_without
number_with

collect.py
```diff
@@ -111,6 +111,10 @@ def do_something(driver, elem_id=None):
             elem.send_keys('prova@email.it')
         elif input_type == 'password':
             elem.send_keys('aMildlyComplexPasswordIn2017')
+        elif input_type == 'checkbox':
```
marco-c: Since this is the same as a `button` or a `a`, can you move it to `if elem.tag_name in ['button', 'a']:`?

marco-c:
Thanks for the PR!

collect.py
sagarvijaygupta: @marco-c the ```tag_name``` for ```checkbox``` is ```input``` so I don't think that we can move it there.

marco-c: Right 😄

PR #18, opened 2018-01-31
Fix all flake8 errors except E722 (bare excepts)

SkullTech:
Fixes all linting errors (flake8) except E722, which prevents bare exceptions. Fixes #16 

I think we should ignore E722 in `setup.cfg`, because as of now, they are unavoidable in some parts of the code.

Issue #6, opened 2018-01-29
Add unit tests for the functions in utils.py


marco-c:

SkullTech:
@marco-c Which testing framework should we use for this project? I suggest [`pytest`](https://docs.pytest.org/en/latest/), it's one of the most robust ones out there.

SkullTech:
I can start working on this once we've decided on the framework to use.

marco-c:
👍 for pytest!

SkullTech:
Okay will start working on it ASAP. :D

marco-c:
Since @SkullTech already has some experience with the project, it's better if someone else who is just starting takes this on to gain some experience.

Shashi456:
Can i take this up ?

marco-c:
Sure, go ahead!

Shashi456:
I wanted to know if these are the following unit tests are okay ? And i'm only majorly checking for functions with some parameters

`def test_mkdir():
    direc_path = os.path.dirname(os.path.realpath(__file__)) + "test"
    mkdir(direc_path)
    assert(os.path.exists(direc_path))`

`
def test_read_labels():
    labels = read_labels(file_name='labels.csv')
    assert('labels' in locals() or 'labels' in globals() )`


`def test_write_labels():
    write_labels(label,file_name='test.csv')
    fname = os.path.dirname(os.path.realpath(__file__)) + "\\test.csv"
    assert(os.path.exists(fname))`

marco-c:
Hey @Shashi456, please write a PR, it's easier to review than a comment in an issue!
For creating directories and files, I suggest you put everything in a temporary directory (you can use the `tempfile` library) and clean everything at the end of the test.

Issue #2, opened 2018-01-29
Label dataset


marco-c:

sagarvijaygupta:
@marco-c Could you please elaborate the issue?

amit2rockon:
labels.json file to be loaded with labeled data information of the data.

amit2rockon:
Some of the data are too explicit.  @marco-c  What to label the data which just differ in single character  'y','d'?

sagarvijaygupta:
@amit2rockon Thanks for the clarification. @marco-c  Should we not let more than one person do the labeling? We will get better voting in that case. Should I right code for that? That might even address the conflicting issue @amit2rockon has mentioned.

marco-c:
I've updated the first comment with a more thorough explanation :)

marco-c:
I've opened #20 for the first point and #21 for the second point. This issue will be about the labeling itself.

Zzoey:
@marco-c A CNN learns more about the patterns in the image (Edges, Corners and their correlations) from example 2 it is evident that it will be difficult for a NN to learn the adversary and classify that both are compatible. 

To detect differences, Y+D and N in a better way or even Y and D+N, I think we can focus more on, Finding ROIs (Attention based) and feed those patches to the NN. This can be our next go-to-go (alternative) if nothing works very well after training part which you suggested.

nok:
At the beginning I would start with screenshots based on equal page sources (same content), so only Y. Furthermore I would try to normalise the device settings to bring the rendered Firefox version closer to the rendered Chrome version. And maybe we could remove the system look and feel elements by injecting a small script before the screenshot will be taken.

marco-c:
> At the beginning I would start with screenshots based on equal page sources (same content), so only Y. 

Do you mean Y vs D+N?
N.B.: The screenshots are always based on equal page sources.

> Furthermore I would try to normalise the device settings to bring the rendered Firefox version closer to the rendered Chrome version. And maybe we could remove the system look and feel elements by injecting a small script before the screenshot will be taken.

This is #13.

nok:
Yes, I meant Y vs D+N.

Thanks, today I don't have an overview over the issues and tasks.

marco-c:
> Thanks, today I don't have an overview over the issues and tasks.

No worries! And keep the suggestions coming ;)
It's better if you file a new issue for every idea you have. In the worst case we close it as a duplicate.

Shashi456:
[...]

PR #19, opened 2018-01-31
removed iterator from utils.py

amit2rockon:


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
+#ignore line limit error
+ignore = E501,E302,E261,E262,E701,E241,E126,E127,E128,E225
```
marco-c: We don't want to ignore all these errors, but only the line limit error for now.

amit2rockon: okay

marco-c:
Thanks for the PR, but #18 preceded you by a few minutes

marco-c:
If you want, you could start working on #6, which is a good first issue.

amit2rockon:
@marco-c  i have labeled #2  around 3000 data, may i pr for it? , but it is saved in json

amit2rockon:
#6 i am trying for it:)

marco-c:
Yes, you can do it, we can transform it to CSV later on.

PR #22, opened 2018-01-31
3000 labelled data in json

amit2rockon:


amit2rockon:
now transforming from json to csv in train.py and label.py

marco-c:
There are several unrelated commits in this PR, can you make a clean one which only updates labels.json?

Issue #20, opened 2018-01-31
Modify label.py and train.py to write/read labels to/from a CSV file instead of a JSON file


marco-c:
This way it is more manageable to see diffs, as each line is a different data point.

marco-c:
Fixed by #28.

PR #23, opened 2018-01-31
Adding 3000 labelled data csv

amit2rockon:


marco-c:
We will need to fix #20 first before merging this.

marco-c:
Sorry @amit2rockon, I've had to revert this PR. CSV and JSON are two different formats, you can't convert between the two just by renaming the file with a different extension.

amit2rockon:
No problem, since labels.py will output csv, but i changed it with excel . currently setting up vm for ubuntu :)

marco-c:
You could convert the json file into csv via a Python script.

amit2rockon:
i tried but output csv file  was same as json file, i work on it as soon as my machine  setup done.

PR #24, opened 2018-01-31
[WIP] Add issue_parser submodule and code to download webcompata data using it

SkullTech:
This code is not working yet because of incompatibility of [issue_parser](https://github.com/webcompat/issue_parser) with Python 3 (see [issue](https://github.com/webcompat/issue_parser/issues/10)). The [issue](https://github.com/webcompat/issue_parser/issues/10) is trying to be fixed by this [pull-request](https://github.com/webcompat/issue_parser/pull/11).


marco-c:
Sounds good to me!
I don't have any preference on where it should be. In tools we have external non-source tools, so I wouldn't put it there. Maybe the top-directory is fine, we don't have too many directories anyway and we can move it later on if we find the need.

marco-c:
@SkullTech do you still intend to work on this?

marco-c:
Closing for lack of response.

Issue #10, opened 2018-01-31
Investigate cases where the crawler wasn't able to take screenshots


marco-c:
There are a few cases where the crawler was not able to take screenshots. We should figure out why and try to fix any issue that we notice.

The files under `data/` are in the format `WEBCOMPAT-ID_ELEMENT-ID_BROWSER.png`. WEBCOMPAT-ID are the IDs from webcompat.com.
ELEMENT-ID are the element IDs where the crawler clicked before taking the screenshot.
BROWSER is the name of the browser.

We should investigate these cases:
1. XXXX_firefox.png is present but XXXX_chrome.png is not present.
2. XXXX_ELEMENT_firefox.png is present but XXXX_ELEMENT_chrome.png is present.

marco-c:
To do this, the first step would be to create a script to list all the inconsistencies.

SkullTech:
I am working on this. Just wanted to let people know so that we don't end up doing duplicate work.

SkullTech:
Which format would be best for exporting the inconsistencies?
- A `CSV` with every row being an inconsistency
- A `JSON` as a list of inconsistencies, the list consisting of dicts detailing it.

marco-c:
Either CSV or a line-limited JSON (https://en.wikipedia.org/wiki/JSON_streaming#Line-delimited_JSON, that is a JSON object per line).

A normal JSON is a bit problematic because you can't easily see diffs between two versions of the file (e.g. if you just add one entry, the diff for a normal JSON file will show you the entire file).

Shashi456:
@marco-c  what do you think we could do next regarding this ?

marco-c:
Manually look at the inconsistencies and see what prevented us from taking a screenshot. E.g. force the crawler to only load the website with an inconsistency and see if the crawler throws an exception in one of the browsers.

Shashi456:
can you point me in a direction as to how i could work with the crawler in this case and force it to load a site?

marco-c:
The crawler is in `collect.py`, you need to change it to load a URL you want instead of loading an URL from one of the webcompat bugs.

marco-c:
We have stored the webcompat ID, so you can retrieve the URLs either with Python by using `utils.get_bugs()` and finding the bug you want, or by loading the bug on the webcompat.com website (e.g. https://webcompat.com/issues/1491).

PR #29, opened 2018-01-31
Investigate inconsistencies in screenshots

SkullTech:
Added a script to extract inconsistencies present in the screenshot archive, and produce a `csv` named `inconsistencies.csv`. The columns of the CSV being  
- WEBCOMPAT-ID
- ELEMENT-ID (Can be blank, representing PNGs of format `XXXX_browser.png`)
- FIREFOX (Presence of Firefox screenshot)
- CHROME (Presence of Chrome screenshot)


SkullTech:
Resolves #10

SkullTech:
Running the script

```console
> python3 data_inconsistencies.py
[*] Getting inconsistencies in screenshots...
[*] 5758 inconsistencies found.
[*] Writing to inconsistencies.csv... Done!
```

Sample lines from the generated `inconsistencies.csv`
```
WEBCOMPAT-ID,ELEMENT-ID,FIREFOX,CHROME
8855,m_login_password_1,True,False
8855,did_submit_4,True,False
8855,login_identify_search_placeholder_5,True,False
8855,did_submit_3,True,False
8855,forgot-password-link_2,True,False
```

marco-c:
I forgot to mention that the actual format of the name also contains a sequence number.
So it is either WEBCOMPAT-ID_ELEMENT-ID_SEQUENCE-NUM_BROWSER.png or WEBCOMPAT-ID_BROWSER.png.

We should sort the CSV results by WEBCOMPAT-ID first and then by SEQUENCE-NUM.

SkullTech:
Done!

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
+import os
+import csv
+
+
+def get_inconsistencies():
+    files = os.listdir('./data/')
+
+    parsed = {}
+    for file in files:
```
marco-c: Can you rename the `file` variable? It's a keyword in Python, so it gets highlighted differently than other variables (which can be confusing).

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def main():
+    print('[*] Getting inconsistencies in screenshots...')
+    incons = get_inconsistencies()
+    print('[*] {} inconsistencies found.'.format(len(incons)))
+
+    print('[*] Writing to inconsistencies.csv... ', end='')
+    with open('inconsistencies.csv', 'w') as csvfile:
+            writer = csv.writer(csvfile, delimiter=',')
```
marco-c: Too many spaces here

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+def get_inconsistencies():
+    files = os.listdir('./data/')
+
+    parsed = {}
+    for file in files:
+        parts = os.path.splitext(file)[0].split('_')
+
+        ID = int(parts[0])
```
marco-c: Same with ID, let's use webcompatID.

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
...
+        if ID not in parsed:
+            parsed[ID] = {}
+        element = '_'.join(parts[1:-1])
+        if element:
```
marco-c: You could avoid this variable here and use the length of parts. If there are two parts, it's the short format, if there are more it's the other format.

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
...
+            spl = element.split('_')
```
marco-c: You can probably reuse `parts` here.

data_inconsistencies.py
```diff
@@ -0,0 +1,54 @@
...
+        parsed[ID][(element, sequence)].append(parts[-1])
+
+    incons = []
+    for key, value in parsed.items():
+        for (element, sequence), browsers in value.items():
+            if len(browsers) < 2:
+                incons.append([key, element, sequence, 'firefox' in browsers, 'chrome' in browsers])
+
+    incons.sort(key=lambda x: x[2])
+    incons.sort(key=lambda x: x[0])
```
marco-c: Nit: you can probably sort in one line if the key lambda returns a tuple instead

data_inconsistencies.py
```diff
@@ -0,0 +1,51 @@
...
+    return incons
+
+
+def main():
+    print('[*] Getting inconsistencies in screenshots...')
+    incons = get_inconsistencies()
+    print('[*] {} inconsistencies found.'.format(len(incons)))
+    print('[*] Writing to inconsistencies.csv... ', end='')
+    with open('inconsistencies.csv', 'w') as csvfile:
+            writer = csv.writer(csvfile, delimiter=',')
```
marco-c: Too much indentation

PR #30, opened 2018-01-31
Add status messages and download progress bar in get_dependencies.py

### LINKED ISSUE #25: Improve user-experience of get_dependencies.py
The `get_dependencies.py` script prints no message, and takes a long time to finish its execution (as it involves downloading and extracting a zip file of size 1.4 GB). So while the script runs, it results in a poor experience of the user (as he doesn't know what's going on, and it gives off the impression that it's stuck). It can be improved by

- Printing some info while the execution occurs.
- Showing a progress bar while it downloads the `data.zip` file.

The progress bar can be added using libraries like [`clint`](https://github.com/kennethreitz/clint), but it also means adding an extra dependency. Although, as we polish this project, we will be needing proper CLI interface, so it won't hurt if we start adding them now.

SkullTech:
Fixes #25 

Added a `download` function in `utils.py`, which takes an URL and a filename as input, and downloads it, while showing a progress bar. Example below

Midway through the operation
```console
λ python3 get_dependencies.py
Using TensorFlow backend.
[*] Extracting webdriver archives...
[*] Downloading data.zip...
[███████████.......................................]
```

### DIFF
--- a/get_dependencies.py
+++ b/get_dependencies.py
@@ -1,5 +1,5 @@
-import sys
 import requests
+import sys
 import os
 import tarfile
 from zipfile import ZipFile
@@ -7,20 +7,20 @@ from zipfile import ZipFile
 
 def download(url, filename):
     with open(filename, 'wb') as f:
-            response = requests.get(url, stream=True)
-            total = response.headers.get('content-length')
+        response = requests.get(url, stream=True)
+        total = response.headers.get('content-length')
 
-            if total is None:
-                f.write(response.content)
-            else:
-                downloaded = 0
-                total = int(total)
-                for data in response.iter_content(chunk_size=max(int(total / 1000), 1024 * 1024)):
-                    downloaded += len(data)
-                    f.write(data)
-                    done = int(50 * downloaded / total)
-                    sys.stdout.write('\r[{}{}]'.format('█' * done, '.' * (50 - done)))
-                    sys.stdout.flush()
+        if total is None:
+            f.write(response.content)
+        else:
+            downloaded = 0
+            total = int(total)
+            for data in response.iter_content(chunk_size=max(int(total / 1000), 1024 * 1024)):
+                downloaded += len(data)
+                f.write(data)
+                done = int(50 * downloaded / total)
+                sys.stdout.write('\r[{}{}]'.format('█' * done, '.' * (50 - done)))
+                sys.stdout.flush()
     sys.stdout.write('\n')
 
 
--- a/get_dependencies.py
+++ b/get_dependencies.py
@@ -1,7 +1,27 @@
+import sys
+import requests
 import os
 import tarfile
 from zipfile import ZipFile
-from utils import download
+
+
+def download(url, filename):
+    with open(filename, 'wb') as f:
+            response = requests.get(url, stream=True)
+            total = response.headers.get('content-length')
+
+            if total is None:
+                f.write(response.content)
+            else:
+                downloaded = 0
+                total = int(total)
+                for data in response.iter_content(chunk_size=max(int(total / 1000), 1024 * 1024)):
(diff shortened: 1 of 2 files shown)


get_dependencies.py
```diff
@@ -1,22 +1,21 @@
 import os
-import shutil
 import tarfile
 from zipfile import ZipFile
-import requests
+from utils import download
```
marco-c: Let's move the `download` function in get_dependencies.py, as it's the only place where we'll use it.

utils.py
```diff
@@ -157,3 +159,22 @@ def balance(it):
 
                 last_label = e[1]
                 yield e
+
+
+def download(url, filename, block='█'):
+    with open(filename, 'wb') as file:
+            response = requests.get(url, stream=True)
```
marco-c: Too many spaces here :)

utils.py
```diff
@@ -157,3 +159,22 @@ def balance(it):
...
+            total = response.headers.get('content-length')
+
+            if total is None:
+                file.write(response.content)
+            else:
+                downloaded = 0
```
marco-c: Nit: No need to define this here

utils.py
```diff
@@ -157,3 +159,22 @@ def balance(it):
 
                 last_label = e[1]
                 yield e
+
+
+def download(url, filename, block='█'):
```
marco-c: Let's remove the `block` parameter and hardcode the character directly in the function.

utils.py
```diff
@@ -157,3 +159,22 @@ def balance(it):
...
+    with open(filename, 'wb') as file:
```
marco-c: Can you use a different name for the `file` variable? It's a Python keyword, so it gets highlighted differently than variables.

utils.py
```diff
@@ -157,3 +159,22 @@ def balance(it):
...
+            response = requests.get(url, stream=True)
+            total = response.headers.get('content-length')
+
+            if total is None:
+                file.write(response.content)
+            else:
+                downloaded = 0
+                total = int(total)
+                for data in response.iter_content(chunk_size=int(total / 1000)):
```
marco-c: Maybe the chunk size is a little too big? If total is 1 GB, the chunk size is 100 MB.

utils.py
SkullTech: I think we do need to define this here to be zero because after that we're going to increment it as

```python
downloaded += len(data)
```

utils.py
SkullTech: I'm dividing the total size by 1000, so for a 1 GB file, it's around 1 MB, which I think is pretty optimum. I went ahead and made it `max(int(total / 1000), 1024 * 1024)`, so that for small files the chunk size doesn't get too small.

utils.py
marco-c: I don't know why I had missed the `+` 😄

utils.py
marco-c: Sounds good. We might also want to introduce a maximum, but it's fine for now.

get_dependencies.py
```diff
@@ -1,22 +1,41 @@
...
 import os
-import shutil
 import tarfile
 from zipfile import ZipFile
-import requests
 
 
+def download(url, filename):
+    with open(filename, 'wb') as f:
+            response = requests.get(url, stream=True)
```
marco-c: Too much indentation here

get_dependencies.py
```diff
@@ -1,22 +1,41 @@
+import sys
+import requests
 import os
-import shutil
 import tarfile
 from zipfile import ZipFile
-import requests
```
marco-c: Nit: move requests back where it was at the beginning.

SkullTech:
Made changes as you recommended. :)

PR #31, opened 2018-01-31
Handle multiple people performing labeling.

### LINKED ISSUE #21: Modify label.py to handle multiple people performing the labeling
Modify the label.py script to handle multiple people performing the labeling (I would suggest simply adding a parameter PERSON, and writing a labels_PERSON.csv according to this parameter). Then, write another script that generates labels.csv from all the labels_*.csv files (basically storing the lines from labels_*.csv which agree on the label).

sagarvijaygupta:
Add UI to take input of filename.
generate_labels.py - creates labels.csv on basis of voting from different files.

Fixes #21


marco-c:
Could you rebase on top of master? There are conflicts in label.py and utils.py.

generate_labels.py
```diff
@@ -0,0 +1,25 @@
...
+
+
+labels_directory = "label_persons/"
+all_file_names = [f for f in listdir(labels_directory)]
+
+labels_voted = {}
+ydn_map = {'y': 0, 'd': 1, 'n': 2}
+ydn_reverse_map = {0: 'y', 1: 'd', 2: 'n'}
+
+for file in all_file_names:
```
marco-c: Can you use a different variable name here? `file` is a Python keyword so it gets highlighted differently than other variable (which can be confusing). Maybe use `file_name`

generate_labels.py
```diff
@@ -0,0 +1,25 @@
...
+    labels = read_labels(labels_directory + file)
+    for key, value in labels.items():
+        if key not in labels_voted.keys():
+            labels_voted[key] = [0, 0, 0]
+        labels_voted[key][ydn_map[value]] += 1
+
+with open('labels.csv', 'w') as f:
```
marco-c: Let's move file writing in utils.py.

label.py
```diff
@@ -58,17 +57,50 @@ def close(e):
...
+        os.makedirs(labels_directory)
+    all_file_names = [f for f in os.listdir(labels_directory)]
+
+    frame = Frame(root)
+    label_file_name = StringVar(frame)
+    frame.pack(side="top")
+    input_type = Label(frame, text="Choose or add filename with .csv extension")
+    input_type.pack(side="top")
+    if len(all_file_names) != 0:
+        old_user = OptionMenu(frame, label_file_name, *all_file_names)
```
marco-c: The option menu is a little confusing, we should add some kind of placeholder.

Or, we can improve the form by doing something like:
Choose existing label file: [OPTIONMENU]
Create new label file: [TEXTINPUT]

Or, we can use argparse and pass the name of the label file as a parameter to the label.py script (so you would run it , for example, as `python label.py marco`) and remove this `choose_file` function.

All three options are fine by me, so feel free to decide which one you prefer.

marco-c:
Another thing, since #23 was merged, please move the current labels.csv file under label_persons (with the name amit2rockon).

sagarvijaygupta:
@marco-c ```labels.csv``` that was merged seems to be a ```json``` file with its extension as ```csv```.

marco-c:
> @marco-c labels.csv that was merged seems to be a json file with its extension as csv.

Thanks for noticing, I will revert the PR.

marco-c:
There's a conflict because I've reverted the PR which was adding the file in the wrong format.

sagarvijaygupta:
@marco-c please review!

label.py
```diff
@@ -1,11 +1,17 @@
-import csv
 from PIL import ImageTk, Image
 from tkinter import Tk, Label
+import argparse
```
marco-c: Nit: move this at the top. For imports we're using the following ordering:
1. Built-in modules first
2. Modules from third-parties
3. Our modules (like `utils`)

sagarvijaygupta: Done!

Issue #32, opened 2018-02-01
Make collect.py support Mac and Windows too


amit2rockon:
Using TensorFlow backend.
10280
_Traceback (most recent call last):
  File "/Users/amit/Documents/GitHub/autowebcompat/collect.py", line 203, in <module>
    firefox_driver = webdriver.Firefox(firefox_profile=firefox_profile, firefox_binary='tools/nightly/firefox-bin')
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/selenium/webdriver/firefox/webdriver.py", line 152, in __init__
    keep_alive=True)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 98, in __init__
    self.start_session(desired_capabilities, browser_profile)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 188, in start_session
    response = self.execute(Command.NEW_SESSION, parameters)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/selenium/webdriver/remote/webdriver.py", line 256, in execute
    self.error_handler.check_response(response)
  File "/Library/Frameworks/Python.framework/Versions/3.6/lib/python3.6/site-packages/selenium/webdriver/remote/errorhandler.py", line 194, in check_response
    raise exception_class(message, screen, stacktrace)
selenium.common.exceptions.WebDriverException: Message: **Unable to find a matching set of capabilities_**

amit2rockon:
Even I tried different versions of selenium and got 3.3.2 to be better , resolved this issue but created the other issue with compatibility of webdrivers.

marco-c:
The script assumes that the system is Linux (it's using the webdriver binaries for Linux and the browser binaries for Linux). We should fix it to support Mac and Windows too.

In the meantime, you can try to run the script in a Linux VM.

marco-c:
Support for Mac has been added with https://github.com/marco-c/autowebcompat/pull/40.

marco-c:
Support for Windows has been added with #42.

Issue #5, opened 2018-01-29
Support 'search' input type in the crawler


marco-c:
https://github.com/marco-c/autowebcompat/blob/e3f77c7b48e017a0ddfe0116e8748964d36fa74c/collect.py#L104

ManasviKundalia:
@marco-c Could you elaborate this issue? I didn't quite get what is to be done here

marco-c:
The crawler currently doesn't support [search input types](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/search). We need to add support for it (close to the line mentioned in the first comment).
It also has to be tested with some website that has a search input type (you can use the MDN docs page).

ManasviKundalia:
@marco-c okay! I will work on it.

PR #34, opened 2018-02-02
* Adding mac support.  *Solving issue #5 support for  'search' input type in the crawler

amit2rockon:
Updated 
`get_dependencies.py`  
 ` collect.py`
`/tools`

changes:-

`get_dependencies.py`  
First it checks the os of the user and then accordingly dowloads the support files for 
* Linux
* Mac os

`collect.py`
* first checks the os of the user and then assign  directory of the bin files of browser accordinlgy
* now it also support search input_type in crawler

`/tools`
* all the support files are now on dropbox hence makes the repository more light .

Till now no support for windows due to non-availability of windows machine but will be added soon.
It solves #32 partially and #5 
i hope this pr will be merged 
if any change  is needed then  please mention


amit2rockon:
Above changes has been tested on both os mac as well as linux and its working accordingly.

amit2rockon:
Now it also solves the issue #5  search input_type  is added.

marco-c:
@amit2rockon each PR should implement a self-contained change. Could you split this PR into two, one for adding Mac support and one for adding support for the `search` input type?

collect.py
```diff
@@ -7,6 +7,18 @@
 from selenium import webdriver
 from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
+from sys import platform as _platform
```
marco-c: This is quite ugly 😄 
Just import `sys` at the top and then use `sys.platform` when you need it.

collect.py
```diff
@@ -7,6 +7,18 @@
...
+if _platform == "linux" or _platform == "linux2":  # linux
+    chromebin = "tools/chrome-linux/chrome"
+    nightybin = 'tools/nightly/firefox-bin'
+
+
+elif _platform == "darwin":    # MAC OS X
+    chromebin = "tools/chrome.app/Contents/MacOS/chrome"
+    nightybin = 'tools/Nightly.app/Contents/MacOS/firefox'
+
+if not os.path.exists('data'):
```
marco-c: Use this snippet instead, which avoids the existance check:
```
    try:
        os.mkdir('data')
    except OSError:
        pass
```

It would be great if you could add a `mkdir` function to utils.py that does this and use it both here and in `utils.prepare_images`.

collect.py
```diff
@@ -115,6 +127,24 @@ def do_something(driver, elem_id=None):
             elem.click()
         elif input_type == 'number':
             elem.send_keys('3')
+        elif input_type == 'search':
+            try:
+                elem.send_keys('marco search')
+                elem.submit()
+            except BaseException:
```
marco-c: We should be more precise with what exception this might throw. What was the exception you've seen?

collect.py
```diff
@@ -115,6 +127,24 @@ def do_something(driver, elem_id=None):
...
+                try:
+                    element = driver.find_element_by_name('search')
```
marco-c: I'm not sure we can rely on the name of the element, here and the following. `q` for example is a pretty generic name, how can we be sure that it is always a search field in any website?

collect.py
```diff
@@ -7,6 +7,18 @@
 from selenium import webdriver
 from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
+from sys import platform as _platform
+if _platform == "linux" or _platform == "linux2":  # linux
+    chromebin = "tools/chrome-linux/chrome"
+    nightybin = 'tools/nightly/firefox-bin'
```
marco-c: Typo, it should be `nightly` and not `nighty`

collect.py
```diff
@@ -7,6 +7,18 @@
 from selenium import webdriver
 from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
+from sys import platform as _platform
+if _platform == "linux" or _platform == "linux2":  # linux
+    chromebin = "tools/chrome-linux/chrome"
```
marco-c: Rename `chromebin` and `nightlybin` to `chrome_bin` and `nightly_bin`.

get_dependencies.py
```diff
@@ -1,15 +1,15 @@
-import sys
 import os
+import sys
```
marco-c: Don't move the import.

get_dependencies.py
```diff
@@ -1,15 +1,15 @@
...
 import tarfile
 from zipfile import ZipFile
 import requests
+from sys import platform as _platform
```
marco-c: Use `sys.platform` when you need it, it's nicer than `_platform`.

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
     sys.stdout.write('\n')
 
 
-print('[*] Extracting webdriver archives...')
-for f in ['geckodriver', 'nightly', 'chromedriver', 'chrome-linux']:
-    tar = tarfile.open('tools/%s.tar.xz' % f, 'r:xz')
-    tar.extractall(path='tools/')
-    tar.close()
+if _platform == "linux" or _platform == "linux2":  # linux
```
marco-c: You should do `platform.startswith('linux')` to detect Linux.

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+    osys = "linux"
+elif _platform == "darwin":    # MAC OS X
+    osys = "mac"
+
+
+if osys == "linux":
+    print('linux os detected !')
```
marco-c: Don't print these logging messages.

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+    print('[*] Downloading support files for linux ~ 108 mb')
```
marco-c: Nit: `MB`

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+    download('https://www.dropbox.com/s/3mosdy9tsf7pilh/linux.zip?dl=1', 'linux.zip')
```
marco-c: Let's try to avoid code duplication, the Linux and Mac code is almost the same, it doesn't make sense to rewrite it twice.

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+    print('[*] Extracting linux.zip...')
+    with ZipFile('linux.zip', 'r') as z:
+        z.extractall("linux")
+    print('[*] Extracting webdriver archives...')
+    for f in ['geckodriver', 'nightly', 'chromedriver', 'chrome-linux']:
+        tar = tarfile.open('linux/%s.tar.xz' % f, 'r:xz')
+        tar.extractall(path='tools/')
+        tar.close()
+    os.remove('linux.zip')
+    os.system("rm -rf linux ")
```
marco-c: We should avoid using `os.system` and use Python standard functions.

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+
+elif osys == "mac":
+    print('mac os detected !')
+    print('[*] Downloading support files for mac ~ 150 mb')
+    download('https://www.dropbox.com/s/k1szymmktj0e1pf/mac.zip?dl=1', 'mac.zip')
+    print('[*] Extracting mac.zip...')
+    os.system("unzip mac.zip -d tools")
```
marco-c: Why aren't you using ZipFile like on Linux?

get_dependencies.py
```diff
@@ -24,18 +24,40 @@ def download(url, filename):
...
+    os.system("cp tools/geckodriver" + " /usr/local/bin/ ")
```
marco-c: There are several problems here:
1. We should avoid os.system and use Python standard functions;
2. Why concatenating two strings here? Why not writing the entire string directly?
3. We should avoid writing any file to `/usr`.

amit2rockon: 1: using python standard functions extracts the file but files get corrupted so used os.system.
2:  yes entire string can be passed directly.
3:  in mac it need to be copied in that directory.

get_dependencies.py
amit2rockon: because files get corrupt after extracting it with standard functions of python

get_dependencies.py
amit2rockon: directory for the bin files in both the cases are different.

get_dependencies.py

collect.py
amit2rockon: okay

get_dependencies.py
amit2rockon: ok

get_dependencies.py
amit2rockon: ok i'l keep in mind

get_dependencies.py
marco-c: This seems really strange. I would expect the standard Python functions to be tested on Mac too. Are the files corrupted without any error message or exception thrown?

get_dependencies.py
marco-c: You can use a variable to store the path of the directory.

amit2rockon:
You may close this PR if it is not so much of importance.   I will work on another issue.

get_dependencies.py
amit2rockon: no error message but the drivers i.e geckodriver and chrome driver get converted to text editable and chrome and nightly does not open .

marco-c:
> You may close this PR if it is not so much of importance. I will work on another issue.

The PR is of importance, but it needs to be split in two PRs, as it is implementing two different things (`Adding mac support.` and `Solving issue #5 support for 'search' input type in the crawler`).

get_dependencies.py
amit2rockon: i used it collect.py ,  and also since 'Google Chrome' contains a space in between therefore it was not working even i  tried /Applications/Google\ Chrome.app/Contents/MacOS/Google\ Chrome but it didn't work , so i  had to change the bin file name to 'chrome' then it started working

get_dependencies.py
marco-c: Maybe you just need to set them as executable?

get_dependencies.py
marco-c: Why does it need to be in /usr/local/bin?

get_dependencies.py
marco-c: I think this is not related to what I asked? I only asked to avoid code duplication between the `linux` and `mac` branches.
E.g. instead of:
```
if linux:
  download(URL_LINUX, NAME_LINUX)
elif mac:
  download(URL_MAC, NAME_MAC)
```
You can do:
```
if linux:
    url = URL_LINUX
    name = NAME_LINUX
elif mac:
    url = URL_MAC
    name = NAME_MAC

download(url, name)
```

get_dependencies.py
amit2rockon: 3. i tested it now and it is actually not required. i did because  earlier i was facing the compability problem so i thought what if put those files into those directory and it started working. i will remove it.

amit2rockon:
How to give split PR ?

get_dependencies.py

marco-c:
> How to give split PR ?

I suggest you learn more about the GitHub flow, which is what we're using for this project (and it is used for most projects on GitHub, so it's going to be useful to learn it anyway).
Here are some resources:
https://guides.github.com/introduction/flow/
https://help.github.com/articles/github-flow/

get_dependencies.py
amit2rockon: then sir, we will have to use one more if else condition for extracting the files.

marco-c: Not if we manage to use ZipFile on Mac too (https://github.com/marco-c/autowebcompat/pull/34#discussion_r165985224).

marco-c:
Closing this as it was superseded by #35.

PR #35, opened 2018-02-06
Adding support for mac

amit2rockon:
#32  is solved , 
* in case of windows , for the first time the admin need to copy the webdrivers to c:\\Windows due to persmission issue, and it is compulsory that webdrivers need to be in that directory .
* in get_dependency.py line 44  no other option found except os.system
@marco-c  review it and mention the changes if needed


collect.py
```diff
@@ -7,6 +8,26 @@
 from selenium import webdriver
 from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
+if sys.platform == "linux" or sys.platform == "linux2":  # linux
```
marco-c: You can use startswith.

get_dependencies.py
```diff
@@ -23,19 +23,39 @@ def download(url, filename):
...
+    print('[*] Extracting win32.zip...')
+    with ZipFile('win32.zip', 'r') as z:
+        z.extractall("tools/")
+    os.remove('win32.zip')
+print('[*] Support files Downloaded!')
 
 print('[*] Downloading data.zip...')
 download('https://www.dropbox.com/s/7f5uok2alxz9j1r/data.zip?dl=1', 'data.zip')
-
-print('[*] Extracting data.zip...')
```
marco-c: Don't remove this message

get_dependencies.py
```diff
@@ -23,19 +23,39 @@ def download(url, filename):
...
+        tar.extractall(path='tools/')
+        tar.close()
+    os.remove('linux.zip')
+    shutil.rmtree('linux')
+
+elif sys.platform == "darwin":
+    print('[*] Downloading support files for mac ~ 150 MB')
+    download('https://www.dropbox.com/s/k1szymmktj0e1pf/mac.zip?dl=1', 'mac.zip')
+    print('[*] Extracting mac.zip...')
+    os.system("unzip mac.zip -d tools")
```
marco-c: We shouldn't use `os.system` here, but the standard ZipFile class. If the extracted executable files can't be run, it's probably just a permission issue that you can solve by setting the executable bits.

get_dependencies.py
```diff
@@ -23,19 +23,39 @@ def download(url, filename):
                 sys.stdout.flush()
     sys.stdout.write('\n')
 
-
-print('[*] Extracting webdriver archives...')
-for f in ['geckodriver', 'nightly', 'chromedriver', 'chrome-linux']:
-    tar = tarfile.open('tools/%s.tar.xz' % f, 'r:xz')
-    tar.extractall(path='tools/')
-    tar.close()
+if sys.platform == "linux":
```
marco-c: sys.platform.startswith('linux')

get_dependencies.py
```diff
@@ -23,19 +23,39 @@ def download(url, filename):
...
+    print('[*] Downloading support files for linux ~ 108 MB')
```
marco-c: The Linux, Windows and Mac code is very similar, we shouldn't replicate it but reuse the code.

get_dependencies.py
amit2rockon: all the files are getting corrupt i am not able to figure out.

utils.py
```diff
@@ -13,6 +13,11 @@ def get_bugs():
     with open('webcompatdata-bzlike.json', 'r') as f:
         return json.load(f)['bugs']
 
+def mk_dir(dir_name):
```
marco-c: Nit: let's call this `mkdir`.

get_dependencies.py
```diff
@@ -23,19 +23,39 @@ def download(url, filename):
...
+    download('https://www.dropbox.com/s/3mosdy9tsf7pilh/linux.zip?dl=1', 'linux.zip')
+    print('[*] Extracting linux.zip...')
+    with ZipFile('linux.zip', 'r') as z:
+        z.extractall("linux")
+    print('[*] Extracting webdriver archives...')
+    for f in ['geckodriver', 'nightly', 'chromedriver', 'chrome-linux']:
```
marco-c: You have removed these files from the repository, so this code should be removed.

marco-c:
We should compress the files with xz and not zip, as the compression ratio is higher.

get_dependencies.py
amit2rockon: actually earlier files are compressed directly and when they are downloaded they get in this form only

utils.py
amit2rockon: okay

get_dependencies.py
marco-c: Try extracting them with ZipFile, then go in the directory where they are extracted, run `ls -al` and post the result here.

get_dependencies.py
marco-c: There's no need to compress them twice then 😄 
Just compress all of the extracted files together in a tar.xz archive.

get_dependencies.py

get_dependencies.py

get_dependencies.py
amit2rockon: ls -al
total 30288
drwxr-xr-x   8 amit  staff       256 Feb  7 00:19 .
drwxr-xr-x  28 amit  staff       896 Feb  7 00:18 ..
-rw-r--r--@  1 amit  staff      6148 Feb  7 00:17 .DS_Store
drwxr-xr-x   3 amit  staff        96 Feb  7 00:19 Nightly.app
drwxr-xr-x   8 amit  staff       256 Feb  7 00:19 __MACOSX
drwxr-xr-x   3 amit  staff        96 Feb  7 00:19 chrome.app
-rw-r--r--   1 amit  staff  11893548 Feb  7 00:19 chromedriver
-rw-r--r--   1 amit  staff   3600548 Feb  7 00:19 geckodriver

amit2rockon: when i try to open the chrome  then a message is displayed : You can’t open the application “chrome” because it may be damaged or incomplete.

marco-c: `geckodriver` and `chromedriver` are not executable, that's probably why you can't run them.

marco-c: Check if `chrome` has the same problem of permissions.

amit2rockon: -rw-r--r--  1 amit  staff  19072 Feb  7 00:19 chrome

marco-c: It does.

amit2rockon: any solution to this?

collect.py
amit2rockon: fine i have done it :)

get_dependencies.py
marco-c: Told you in the first comment 😄 
Try to set the files as executable.

marco-c:
> in case of windows , for the first time the admin need to copy the webdrivers to c:\Windows due to persmission issue, and it is compulsory that webdrivers need to be in that directory .

I don't think this is actually needed. You probably just need to add the webdrivers to your path.

amit2rockon:
no it was not working.

marco-c:
As I told you in #34, each PR should be a self-contained change. Let's add support for Mac first and then for Windows in another PR.

amit2rockon:
okay :)

marco-c:
> no it was not working.

How did you add the webdrivers to your path?

get_dependencies.py
amit2rockon: okay 😄

amit2rockon:
i had to manually keep the files inC:\\windows  ,it asked for administration permission and with a "ok" i did it, i tried it with python but getting error  .

marco-c:
> i had to manually keep the files inC:\windows ,it asked for administration permission and with a "ok" i did it, i tried it with python but getting error .

OK, so you did not add the webdrivers to your path. Here's a description of what the PATH is: https://en.wikipedia.org/wiki/PATH_(variable).
You'll probably just need to change https://github.com/marco-c/autowebcompat/blob/master/collect.py#L196 to use the correct path separator (`os.pathsep`) instead of `:` (which only works on Linux and Mac).

PR #36, opened 2018-02-06
[WIP] Write unit tests for the functions in utils.py

SkullTech:
This is a WIP as of now. Made pull request to get input from project maintainers. @marco-c

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,15 @@
+import sys
+import os
+sys.path.append(os.path.dirpath(os.path.realpath(__file__)) + '/../')
+
+import utils  # noqa: E402
+
+
+def test_get_bugs():
+    bugs = utils.get_bugs()
+    assert bugs
+
+
+def test_get_all_images():
+    images = utils.get_all_images()
+    assert images


tests/test_utils.py
```diff
@@ -0,0 +1,15 @@
+import sys
+import os
+sys.path.append(os.path.dirpath(os.path.realpath(__file__)) + '/../')
```
SkullTech: Is this an acceptable way to import the utils.py module from the outer directory? Seems like a hack to me, but this is what I got after some stackoverflow searches.

marco-c: Can we avoid this? Maybe we can restructure the project like the first example at https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code.

marco-c: I hadn't seen this comment 😄 
I've replied at https://github.com/marco-c/autowebcompat/pull/36#discussion_r166479187.

SkullTech: The first example's directory structure looks good. If we rearrange our project like that, will we able to import the project modules directly? like
```python
```

SkullTech:
Created issue #37 in reference to this.

tests/test_utils.py
marco-c: I think so, if you execute `python3 -m pytest` or `PYTHONPATH=. pytest`.

Instead of restructuring, we can also run `PYTHONPATH=.. pytest`.

SkullTech: I think restructuring will also help us in the long run. So better to just do it now.

SkullTech: What do you think?

marco-c: I'm not totally sure because the source files are executable scripts rather than modules that you import.
Only `network.py` and `utils.py` are actually modules.
Maybe we can restructure putting `network.py` and `utils.py` in a `autowebcompat` directory, but keeping the other scripts where they are.

Zzoey: @SkullTech @marco-c  I think its always better to avoid system calls. :) Once the structure of the directory is set, it will be easier. OOP structure favours imports with ease.

Can I structure the whole network.py into a nice class with appropriate functions ? What do you think ?

marco-c: > @SkullTech @marco-c I think its always better to avoid system calls. :) Once the structure of the directory is set, it will be easier. OOP structure favours imports with ease.

For now, we are leaning towards moving the `network.py` and `utils.py` modules into a `autowebcompat` directory and leaving the scripts (which are never used as modules) in the top-level directory. Discussion at https://github.com/marco-c/autowebcompat/pull/36#pullrequestreview-94547867.

> Can I structure the whole network.py into a nice class with appropriate functions ? What do you think ?

Sounds good to me, this should happen regardless of the directory structure.

marco-c:
Hey @SkullTech, since #6 is a "good first issue", I'll close this and let someone without experience with the project take it.

PR #39, opened 2018-02-07
VGG_16 Model Added along with Binary CrossEntropy Loss.

Zzoey:


marco-c:
Thanks @Zzoey, there are some linting issues that need to be fixed.

I think we should find a nice way for users of the `network.py` module to choose the network and optimizer they want to use.
For example, we could add a `network` parameter to the `create` function and make it call the correct function according to that:
```
def create(input_shape, network='vgglike'):
    network_func = globals()['create_%s_network' % network]
    base_network = network_func(input_shape)
```

Zzoey:
@marco-c Yeah. That seems nice. I have changed it accordingly also added to use weights in our siamese module. Please let me know if you have suggestions.

network.py
```diff
@@ -49,12 +88,17 @@ def create_vgglike_network(input_shape):
...
-def create(input_shape):
-    base_network = create_vgglike_network(input_shape)
+def create(input_shape, network='vgglike', weights=None):
+    network_func = globals()['create_%s_network' % network]
+    base_network = network_func(input_shape)
 
     input_a = Input(shape=input_shape)
     input_b = Input(shape=input_shape)
 
+    # Loading pretrained weights corresponsing to the network used
```
marco-c: Typo `corresponsing` -> `corresponding`

network.py
```diff
@@ -85,6 +129,10 @@ def accuracy(y_true, y_pred):
 
 def compile(model):
```
marco-c: We might want to do the same here, to allow choosing the optimizer, but we can do it in a follow-up PR.

marco-c:
Thanks, it looks good. Just a typo to fix.

network.py
Zzoey: @marco-c changed the compile function a bit more. Hope it looks good. :)

network.py
```diff
@@ -83,8 +127,14 @@ def accuracy(y_true, y_pred):
...
-    opt = RMSprop()
-    # opt = SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True)
-    # opt = Nadam()
-    model.compile(loss=contrastive_loss, optimizer=opt, metrics=[accuracy])
+def compile(model, optimizer=None, loss_func=False):
+
+    allOptimizers = {'sgd': SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True),
+        'adam': Adam(), 'nadam': Nadam(), 'rms': RMSprop()}
+
+    if optimizer in allOptimizers:
```
marco-c: Let's turn this into an `assert`, so we ensure we always pass a valid optimizer.

network.py
```diff
@@ -83,8 +127,14 @@ def accuracy(y_true, y_pred):
...
 
 
-def compile(model):
-    opt = RMSprop()
-    # opt = SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True)
-    # opt = Nadam()
-    model.compile(loss=contrastive_loss, optimizer=opt, metrics=[accuracy])
+def compile(model, optimizer=None, loss_func=False):
+
+    allOptimizers = {'sgd': SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True),
```
marco-c: Nit: I'd prefer the style:
```Python
allOptimizers = {
    'sgd': ...,
    'adam': ...,
    ...
}
```

network.py
```diff
@@ -83,8 +127,14 @@ def accuracy(y_true, y_pred):
     return K.mean(K.equal(y_true, K.cast(y_pred < 0.5, y_true.dtype)))
 
 
-def compile(model):
-    opt = RMSprop()
-    # opt = SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True)
-    # opt = Nadam()
-    model.compile(loss=contrastive_loss, optimizer=opt, metrics=[accuracy])
+def compile(model, optimizer=None, loss_func=False):
```
marco-c: Instead of `False`, we can use `contrastive_loss` as the default directly, so you can also remove the `loss_func = contrastive_loss if loss_func == False else 'binary_crossentropy'` line.

network.py
```diff
@@ -83,8 +127,14 @@ def accuracy(y_true, y_pred):
...
+
```
marco-c: Nit: no newline here

network.py
```diff
@@ -83,8 +127,15 @@ def accuracy(y_true, y_pred):
...
-    model.compile(loss=contrastive_loss, optimizer=opt, metrics=[accuracy])
+def compile(model, optimizer=None, loss_func=contrastive_loss):
+    allOptimizers = {
+        'sgd': SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True),
+        'adam': Adam(),
+        'nadam': Nadam(),
+        'rms': RMSprop()
+    }
+
+    if optimizer in allOptimizers:
```
marco-c: Last change, assert that this is true, so we fail clearly when the optimizer is not in the list of supported ones.
```Python
assert optimizer in allOptimizers, '%s is an invalid optimizer' % optimizer
```

Zzoey: Okay Sure. :) Yea. Thanks. Next time I will keep track of these.

network.py
Zzoey: sorry I missed this comment. Apologies!

Issue #15, opened 2018-01-31
Investigate if we can repeat the same steps in different browsers without element IDs


marco-c:
Right now, we only interact with elements that have an ID and store the sequence of operations as a list of element IDs.
A lot of times on websites the elements don't have an ID though. Can we find a way to repeat the same steps that we perform in a browser on another browser without having the IDs of the elements?

sagarvijaygupta:
@marco-c  we can try using the absolute ```XPath```. I think it should work.

marco-c:
> @marco-c we can try using the absolute XPath. I think it should work.

We can try, the only problem is that I'm not sure it's supported exactly in the same way across browsers.

sagarvijaygupta:
@marco-c I was also searching about that only. It feels it won't be that feasible.

Though I found a much **better idea**!
 Considering we have all children with which we feel we can interact with, we can store all the attributes of that child we wish to work with even if it doesn't have ID. 
So keeping this sequence saved we can than search all the elements in the next browser and match with all the attributes. 
If they come out to be same we follow the sequence.
Below is the code to get all the attributes of element - child

` all_attributes_of_child = driver.execute_script
('
var elem_attribute = {};
for (i = 0; i < arguments[0].attributes.length; i++) 
 { 
         elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value
 }; 
         return  elem_attribute;
', child)`

This file contains output for both firefox and chrome of all attributes. Please take a look and tell me if this idea is feasible? 
[firefox_chrome.txt](https://github.com/marco-c/autowebcompat/files/1687265/firefox_chrome.txt)

sagarvijaygupta:
@marco-c any thoughts on this?

sagarvijaygupta:
@marco-c  any thoughts on this?

sagarvijaygupta:
@marco-c  Please give a review on this idea? I wish to work on it.

marco-c:
The idea sounds good to me. I think we should keep both the ID version and add this capability too.
We should create a file alongside the screenshot to tell how the screenshot was generated (so that we can re-generate it if needed). Also the naming of the screenshots needs to change, as it currently contains the IDs.
It would be great if you could rewrite that script in Python, but doing it on the JavaScript side is fine too.

marco-c:
The screenshot naming could be:
WEBCOMPAT-ID_BROWSER.png for the first load screenshots;
[...]

PR #41, opened 2018-02-07
Add repetition of steps in different browsers without element ids.

sagarvijaygupta:
Solves #15


sagarvijaygupta:
@marco-c please review it.

collect.py
```diff
@@ -89,21 +90,20 @@ def do_something(driver, elem_id=None):
...
                 continue
 
-            elem = child
-            break
+            if elem_attributes is None or elem_attributes == child_attributes:
+                elem = child
+                break
     else:
+        elem_id = elem_attributes['id']
         elem = driver.find_element_by_id(elem_id)
```
marco-c: This is still only using the ID. It means we will find many more possible elements in the first browser, but then won't be able to reproduce the same steps in the second browser.

marco-c:
I'm thinking that to avoid confusion, maybe we can actually only have two possible formats:
WEBCOMPAT-ID_BROWSER.png for the first load screenshots;
WEBCOMPAT-ID_SEQUENCE-NUM_BROWSER.png for the screenshots generated after interacting with an element (and a single WEBCOMPAT-ID_SEQUENCE.txt file containing the attributes of the elements interacted with).

And we have to convert the already existing data to this format.
To do it, we should write a script that renames the WEBCOMPAT-ID_ELEMENT-ID_SEQUENCE-NUM_BROWSER.png files to WEBCOMPAT-ID_SEQUENCE-NUM_BROWSER.png, storing the ELEMENT-IDs in a WEBCOMPAT-ID_SEQUENCE.txt file.

collect.py
sagarvijaygupta: @marco-c We are able to reproduce same steps in second browser as in line 80 I do check if the ```id``` is present in ```elem_attributes```. In most cases it is not found so the code enters the ```if``` condition and finds the element on the basis of attributes. Images in #15 demonstrate how the code reproduces the same steps for the second browser.

marco-c: Oh you're right, sorry I hadn't noticed.
I would refactor this a bit so that the first branch is taken when we are looking for an element to choose, and the second branch when we already know the element we want to use (either with the attributes or the ID).

sagarvijaygupta: @marco-c  I didn't understand what do you mean by refactoring it? Should I create two functions?

marco-c: I mean something like this:
```Python
if elem.attributes is None:
    CODE THAT CHOOSES AN ELEMENT FROM THE PAGE
else:
    FIND AN ALREADY CHOSEN ELEMENT BY ID OR USING ALL THE ATTRIBUTES
```

So the first branch implements the "choose random element" part, the second branch implements the "I already know what element I want to use, let's find it on the page" part.

sagarvijaygupta: @marco-c  I made the required change. Please check.

collect.py
```diff
@@ -92,22 +93,43 @@ def do_something(driver, elem_id=None):
...
 
         for child in children:
-            elem_id = child.get_attribute('id')
 
-            # We need to store the ID in order to replicate what we are doing, so we
-            # have to skip elements with no ID.
-            if elem_id == '':
-                continue
+            # Get all the attributes of the child.
+            child_attributes = driver.execute_script('var elem_attribute = {}; for (i = 0; i < arguments[0].attributes.length; i++) { elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value }; return elem_attribute;', child)
```
marco-c: Can you make this script more readable by using a multiline string? Like in `wait_loaded`.

marco-c: It would be great if we could use the Selenium API to do the same, but for now it's fine.

collect.py
```diff
@@ -92,22 +93,43 @@ def do_something(driver, elem_id=None):
...
+
+            buttons = body.find_elements_by_tag_name('button')
+            links = body.find_elements_by_tag_name('a')
+            inputs = body.find_elements_by_tag_name('input')
+            children = buttons + links + inputs
+
+            for child in children:
+
+                # Get all the attributes of the child.
+                child_attributes = driver.execute_script('var elem_attribute = {}; for (i = 0; i < arguments[0].attributes.length; i++) { elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value }; return elem_attribute;', child)
```
marco-c: Since this is repeated, it should be in a separate function.

collect.py
```diff
@@ -92,22 +93,43 @@ def do_something(driver, elem_id=None):
...
-            if elem_id == '':
-                continue
+            # Get all the attributes of the child.
+            child_attributes = driver.execute_script('var elem_attribute = {}; for (i = 0; i < arguments[0].attributes.length; i++) { elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value }; return elem_attribute;', child)
 
             # If the element is not displayed or is disabled, the user can't interact with it. Skip
             # non-displayed/disabled elements, since we're trying to mimic a real user.
             if not child.is_displayed() or not child.is_enabled():
                 continue
-
```
marco-c: Don't remove this newline

collect.py
```diff
@@ -92,22 +93,43 @@ def do_something(driver, elem_id=None):
...
+            links = body.find_elements_by_tag_name('a')
+            inputs = body.find_elements_by_tag_name('input')
+            children = buttons + links + inputs
+
+            for child in children:
+
+                # Get all the attributes of the child.
+                child_attributes = driver.execute_script('var elem_attribute = {}; for (i = 0; i < arguments[0].attributes.length; i++) { elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value }; return elem_attribute;', child)
+
+                # If the element is not displayed or is disabled, the user can't interact with it. Skip
```
marco-c: I think we can ignore this part, as we already know that the element was displayed and not enabled in the first browser.

collect.py
```diff
@@ -197,6 +220,10 @@ def run_tests(firefox_driver, chrome_driver):
                not os.path.exists('data/%d_chrome.png' % bug['id']):
                 sequence = run_test(bug, 'firefox', firefox_driver)
                 run_test(bug, 'chrome', chrome_driver, sequence)
+
+                with open("data/" + str(bug['id']) + ".txt", 'w') as f:
+                    f.write(json.dumps(sequence))
```
marco-c: Let's write an object of the sequence per line, instead of a full json object. This way it's easier to see diffs.

collect.py
```diff
@@ -161,18 +183,19 @@ def run_test(bug, browser, driver, op_sequence=None):
...
             else:
-                elem_id = op_sequence[i]
-                do_something(driver, elem_id)
+                elem_attributes = op_sequence[i]
+                do_something(driver, elem_attributes)
 
-            print('  - Using %s' % elem_id)
+            print('  - Using %s' % elem_attributes)
+            image_file = str(bug['id']) + '_' + str(i) + '_' + browser
+            screenshot(driver, 'data/%s.png' % (image_file))
```
marco-c: Since this is breaking compatibility with the old way of doing things, before merging this we will also need a script to convert the current data to the new format.

collect.py
sagarvijaygupta: @marco-c  I will try to do it with Selenium API.

collect.py
sagarvijaygupta: Done.

collect.py
sagarvijaygupta: @marco-c  Because of two browsers can display in different way it might be that one browser displays the element while other doesn't. Should I remove it?

collect.py
sagarvijaygupta: Ok!

collect.py
sagarvijaygupta: Done.

collect.py
```diff
@@ -76,10 +77,24 @@ def close_all_windows_except_first(driver):
     driver.switch_to_window(windows[0])
 
 
-def do_something(driver, elem_id=None):
+def get_all_attributes(driver, child):
+    child_attributes = driver.execute_script("""
+        var elem_attribute = {};
```
marco-c: Use `let` instead of `var`.

collect.py
```diff
@@ -76,10 +77,24 @@ def close_all_windows_except_first(driver):
...
+
+        for (i = 0; i < arguments[0].attributes.length; i++) {
```
marco-c: Use `let i`

collect.py
```diff
@@ -197,6 +235,11 @@ def run_tests(firefox_driver, chrome_driver):
                not os.path.exists('data/%d_chrome.png' % bug['id']):
                 sequence = run_test(bug, 'firefox', firefox_driver)
                 run_test(bug, 'chrome', chrome_driver, sequence)
+
+                with open("data/" + str(bug['id']) + ".txt", 'w') as f:
```
marco-c: Nit: Use `%` instead of `+` as it is more readable

`'data/%d.txt' % bug['id']`

collect.py
```diff
@@ -107,7 +119,32 @@ def do_something(driver, elem_id=None):
...
+            assert len(body) == 1
+            body = body[0]
+
+            buttons = body.find_elements_by_tag_name('button')
+            links = body.find_elements_by_tag_name('a')
+            inputs = body.find_elements_by_tag_name('input')
+            children = buttons + links + inputs
+
+            for child in children:
+
```

collect.py
marco-c: If that happens, it's a compatibility issue and we want to detect it, so it's useful to take a screenshot even if the element is not displayed in the second browser.

collect.py
```diff
@@ -92,12 +107,9 @@ def do_something(driver, elem_id=None):
         random.shuffle(children)
 
         for child in children:
-            elem_id = child.get_attribute('id')
```

rename_images.py
```diff
@@ -0,0 +1,19 @@
+from os import listdir, rename
```
marco-c: The script should also generate the `.txt` files accompanying the images.
I suggest you use `split('_')` to get the components from the name (if there are two components, it's an image in the format WEBCOMPAT-ID_BROWSER.png and you can skip it; if there are more than two components, it's an image in the format WEBCOMPAT-ID_ELEMENT-ID_SEQUENCE-NUM_BROWSER.png and you have to rename it and put in the `.txt` file in order of SEQUENCE-NUM). You can use this https://github.com/marco-c/autowebcompat/blob/master/data_inconsistencies.py#L12 as inspiration

sagarvijaygupta: Done!

collect.py
```diff
@@ -90,14 +105,10 @@ def do_something(driver, elem_id=None):
         children = buttons + links + inputs
 
         random.shuffle(children)
-
         for child in children:
-            elem_id = child.get_attribute('id')
```
marco-c: Nit: remove newline here

collect.py
```diff
@@ -107,7 +118,31 @@ def do_something(driver, elem_id=None):
...
+            body = driver.find_elements_by_tag_name('body')
+            assert len(body) == 1
+            body = body[0]
+
+            buttons = body.find_elements_by_tag_name('button')
+            links = body.find_elements_by_tag_name('a')
+            inputs = body.find_elements_by_tag_name('input')
+            children = buttons + links + inputs
+            for child in children:
+
```
marco-c: Nit: remove newline here

collect.py
```diff
@@ -107,7 +118,31 @@ def do_something(driver, elem_id=None):
...
+                # Get all the attributes of the child.
+                child_attributes = get_all_attributes(driver, child)
+
+                # If the element is not displayed or is disabled, the user can't interact with it. Skip
+                # non-displayed/disabled elements, since we're trying to mimic a real user.
+                if not child.is_displayed() or not child.is_enabled():
+                    continue
```
marco-c: As discussed, this part should be removed.

rename_images.py
```diff
@@ -0,0 +1,20 @@
+import os
+
+files = os.listdir('./data/')
+
+image_info = {}
+for f in files:
+    parts = os.path.splitext(f)[0].split('_')
+    if len(parts) <= 2:
+        continue
+    if parts[0] not in image_info.keys():
```
marco-c: Assign names to the parts before this line, otherwise the rest of the code is hard to read.

rename_images.py
```diff
@@ -0,0 +1,20 @@
...
+        image_info[parts[0]] = {}
+    image_info[parts[0]][parts[-2]] = parts[1:-2]
+    new_name = "%s_%s_%s.png" % (parts[0], parts[-2], parts[-1])
+    os.rename("data/" + f, "data/" + new_name)
+
+for key, attributes in image_info.items():
+    with open("./data/%s.txt" % key, "w") as text_file:
+        attributes = sorted(attributes.items())
```
marco-c: Isn't this going to sort by element ID rather than sequence number?

sagarvijaygupta: @marco-c No actually because ```attributes``` is a ```dictionary``` of ```{ seq_no : elem_id }```
I changed the variables for better understanding.

collect.py
sagarvijaygupta: Oh ok! I get it.

collect.py
```diff
@@ -107,7 +117,24 @@ def do_something(driver, elem_id=None):
...
+        if 'id' not in elem_attributes.keys():
+            body = driver.find_elements_by_tag_name('body')
+            assert len(body) == 1
+            body = body[0]
+
+            buttons = body.find_elements_by_tag_name('button')
+            links = body.find_elements_by_tag_name('a')
+            inputs = body.find_elements_by_tag_name('input')
+            children = buttons + links + inputs
+            for child in children:
```
marco-c: Nit: newline before this for

rename_images.py
```diff
@@ -0,0 +1,26 @@
...
+    image_info[webcompat_id][sequence_no] = elem_id
+    new_name = "%s_%s_%s.png" % (webcompat_id, sequence_no, browser)
+    os.rename("data/" + f, "data/" + new_name)
+
+for webcompat_id, seq_no_and_elem_id in image_info.items():
+    with open("./data/%s.txt" % webcompat_id, "w") as text_file:
+        seq_no_and_elem_id = sorted(seq_no_and_elem_id.items())
+        for value in seq_no_and_elem_id:
+            sequence_no = value[0]
+            elem_id = value[1][0]
```
marco-c: This is going to use only the first part of the element ID (e.g. for `an_element_id`, this variable would be `an`). I would replace line 12 with `'_'.join(parts[1:-2])` and this line with `value[1]`.

collect.py
```diff
@@ -90,14 +104,10 @@ def do_something(driver, elem_id=None):
         children = buttons + links + inputs
 
         random.shuffle(children)
-
```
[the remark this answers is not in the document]
sagarvijaygupta: @marco-c Is there any way to know where to put newline and where not? :laughing:

rename_images.py
sagarvijaygupta: @marco-c Done!

collect.py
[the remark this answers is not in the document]
marco-c: 1) Don't remove newlines unless they are related to your changes;
2) No newlines after a if, for, def, etc.;
3) Newline to separate two logic blocks.

rename_images.py
```diff
@@ -0,0 +1,26 @@
...
+    new_name = "%s_%s_%s.png" % (webcompat_id, sequence_no, browser)
+    os.rename("data/" + f, "data/" + new_name)
+
+for webcompat_id, seq_no_and_elem_id in image_info.items():
+    with open("./data/%s.txt" % webcompat_id, "w") as text_file:
+        seq_no_and_elem_id = sorted(seq_no_and_elem_id.items())
+        for value in seq_no_and_elem_id:
+            sequence_no = value[0]
+            elem_id = value[1]
+            text_file.write("%s %s\n" % (sequence_no, elem_id))
```
marco-c: Note that here you're writing text, but in `collect.py` you are writing json objects.
The format of the files should be the same.

sagarvijaygupta: @marco-c For format to be same I made ```{"id" : "elem_id"}``` per line according to sequence number.

collect.py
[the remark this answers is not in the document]
sagarvijaygupta: @marco-c I will make sure nothing related to newline comes in my future PRs

rename_images.py
marco-c: It would be better to do `json.dumps({'id': elem_id})`

marco-c: It doesn't matter though, the end result is the same.

Can you run the script on your machine to rename the files? Then zip the resulting `data` directory and upload it on Dropbox or some other service. Before merging this PR, I will have to update `get_dependencies.py` to use the new data.zip file.

collect.py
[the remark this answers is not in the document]
marco-c: Don't worry, they are minor annoyances.

collect.py
marco-c: This is still here 😄

collect.py
marco-c: This is still here 😄

rename_images.py
sagarvijaygupta: @marco-c  I will try to upload it today. Got a better Internet connection :smile:

sagarvijaygupta: @marco-c 
https://www.dropbox.com/s/rsdq581uteu3mtc/data.zip?dl=0 
this is the link with the renamed files.

sagarvijaygupta:
@marco-c I just realized the ```assert``` statement in ```get_images``` in ```utils.py``` called in ```labels.py``` will fail. Should I change it into an ```if``` statement?

marco-c:
> @marco-c I just realized the assert statement in get_images in utils.py called in labels.py will fail. Should I keep the txt files in another folder?

We should update `get_all_images` in utils.py to only return png files, and update `prepare_images` to use `get_all_images` instead of `os.listdir`.

The rename_images.py script should also update the labels files.

marco-c:
> We should update get_all_images in utils.py to only return png files, and update prepare_images to use get_all_images instead of os.listdir.

I've done these changes myself.

The only thing missing now is rewriting the label files with the new image names. Once that's done, I will replace the current data.zip file with the one you uploaded, then you can remove the rename_images.py script from the PR and we can finally merge it!

sagarvijaygupta:
@marco-c I made the required changes. (removed ```rename_images.py``` and updated the ```labels.csv``` and other csv)

PR #42, opened 2018-02-08
Adding support for windows 

amit2rockon:
@marco-c  review it


collect.py
```diff
@@ -15,6 +15,11 @@
     chrome_bin = "tools/chrome.app/Contents/MacOS/chrome"
     nightly_bin = 'tools/Nightly.app/Contents/MacOS/firefox'
```

collect.py
```diff
@@ -15,6 +15,11 @@
...
 
+elif sys.platform.startswith("win32"):
+    path = os.getcwd()
```
marco-c: Are you sure this is needed?

get_dependencies.py
```diff
@@ -31,6 +31,11 @@ def download(url, filename):
     url = 'https://www.dropbox.com/s/k4yifantsypy9xv/mac.tar.xz?dl=1'
     name = 'mac.tar.xz'
 
+elif sys.platform.startswith('win32'):
+    url = 'https://www.dropbox.com/s/scmgetma3d8jtl6/win32.tar.xz?dl=1'
```
marco-c: Use `https://www.dropbox.com/s/xskj9rpn2fjkra8/win32.tar.xz?dl=1` (uploaded on my Dropbox)

collect.py
amit2rockon: i thought it will need the full directory. okay, i will confirm it, currently, i don't have the machine now.

get_dependencies.py
```diff
@@ -31,6 +31,11 @@ def download(url, filename):
     url = 'https://www.dropbox.com/s/k4yifantsypy9xv/mac.tar.xz?dl=1'
     name = 'mac.tar.xz'
```
marco-c: Nit: no newline here

collect.py
amit2rockon: yes sir, it is required. without it, compability errors are shown.

amit2rockon:
@marco-c  please review it.

collect.py
marco-c: What errors?

amit2rockon: https://pastebin.mozilla.org/9077598

marco-c: What is the value of nightly_bin that triggers this error?

amit2rockon: How to check this on windows machine ?

amit2rockon: and it starts functioning normally when path is added.

amit2rockon: '\\tools\\Nightly\\firefox.exe'

amit2rockon: its "\\\"

amit2rockon: actually its two times \

marco-c: Have you tried using a relative path? Without the initial `\` before tools

amit2rockon: still it is present even without initial \\

amit2rockon: means the error still comes.
errorr occurs when i use \\tools\\Nightly\\firefox.exe
but not in case of  os.getcwd() + \\tools\\Nightly\\firefox.exe
what is the downside of using os.getcwd?

amit2rockon: apologies me , its working 😃 now

amit2rockon: i ran  the script after rebooting the machine and its working.

amit2rockon:
:)

Issue #47, opened 2018-02-09
Make `download` function in get_dependencies.py automatically detect the file name from the URL


marco-c:

rahuldcoder:
Is this issue fixed??I was looking for good first issues on this repository.

marco-c:
It's not fixed yet, but there's already a PR that is trying to implement this.

marco-c:
It's not fixed yet, but there's already a PR (#48) that is trying to implement this.

PR #48, opened 2018-02-09
fixing #47 filename finder in  download get_dependency.py

amit2rockon:
solving #47


get_dependencies.py
```diff
@@ -5,7 +5,8 @@
 import requests
 
 
-def download(url, filename):
+def download(url):
+    filename = url[url.rfind("/") + 1:url.find("?")]
```
marco-c: This is a bit brittle and could break if we start using other URLs.
We should use the `Content-Disposition` header of the response and, if it doesn't exist, use the basename of the URL.

get_dependencies.py
```diff
@@ -26,17 +27,14 @@ def download(url, filename):
...
-    name = 'mac.tar.xz'
 elif sys.platform.startswith('win32'):
     url = 'https://www.dropbox.com/s/xskj9rpn2fjkra8/win32.tar.xz?dl=1'
-    name = 'win32.tar.xz'
 
 print('[*] Downloading support files...')
-download(url, name)
-
+download(url)
+name = url[url.rfind("/") + 1:url.find("?")]
```
marco-c: Make `download` return the name, so we avoid duplicating the code here.

get_dependencies.py
amit2rockon: what about the case when the url will not contain the file name?

get_dependencies.py
amit2rockon: fname = download(url)
 tarfile.open(fname, 'r:xz')
....
os.remove(fname)
may i use this way if download return the filename?

get_dependencies.py
marco-c: if there's no `Content-Disposition` header and the basename is an empty string, throw an exception (`Couldn't get file name for this URL`, or something like this).

get_dependencies.py
marco-c: Why do you want to change the name of the variable from `name` to `fname`?

get_dependencies.py

get_dependencies.py
amit2rockon: variable will be name only , then is this permissible  assigning download(url) to name?

amit2rockon: apologies me , but just asking that download(url) will download file though it will return filename yet assigning to a variable named 'name' may cause to misunderstand the variable (for the new ones).

marco-c: I still don't understand the question 😄 
Let's leave the name of the variable as `name`, we can consider changing it in the future in a separate PR.

amit2rockon:
@marco-c  review it.

get_dependencies.py
```diff
@@ -5,9 +5,11 @@
 import requests
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = response.headers.get('content-disposition')
+    filename = filename[filename.find("\"") + 1: filename.rfind("\"")]
```
marco-c: This is not implementing the fallback when the content-disposition header doesn't exist.
I suggest you use the `rfc6266` library to do this.

konemshad: @marco-c   i think the pr of mine which you closed recently implements your requirement, doesn't it?

marco-c: I don't think so, in your PR you're printing `File Not Found` when the content-disposition header doesn't exist and the `filename` variable is not defined in that case.

konemshad: OK. what should we store the value in the variable "filename" if the content-disposition header does not exist ? we can just add an else clause?

marco-c: Discussion at https://github.com/marco-c/autowebcompat/pull/48#discussion_r167229099.

ManasviKundalia:
[@marco-c is this okay? ](https://github.com/ManasviKundalia/autowebcompat/blob/c48b9d2693ea6e0ff175792837a111a13d8af701/get_dependencies.py)

amit2rockon:
okay

marco-c:
> @marco-c is this okay? 

There are some problems with your solution:
1) It's using filename_unsafe, which is, as the name says, unsafe 😄 
2) It's not implementing the fallback of using the basename of the file from the URL when the header doesn't exist;
3) I would just use parse_requests_response directly

ManasviKundalia:
@marco-c I have made the changes [here](https://github.com/ManasviKundalia/autowebcompat/blob/ManasviKundalia-patch-1/get_dependencies.py) 
Do you have any examples/ corner cases on which I could test the code?

amit2rockon:
@marco-c  can it be generalized for every link even if the link is containing the downloadable file?
like this link.    
https://drive.google.com/uc?export=download&confirm=oIyQ&id=17sqc0GQlUiFHej-OZPasJS5ionZkoCK1
though in the case of google drive google adds BOM (Byte Order Marks) to the beginning of some files.
 and this link contains linux.zip (old file) even I tried @ManasviKundalia last committed code and it is actually getting failed at this point.
so if the file is going to be only on dropbox then its okay but what if such link is present?

amit2rockon:
@marco-c  where to upload the downloaded extra screenshots? 
and what if other's downloaded screenshots will have the same name but different files?

marco-c:
> @marco-c can it be generalized for every link even if the link is containing the downloadable file?

I don't understand this question, sorry. Could you rephrase?
Maybe you mean what should we do when the `Content-Disposition` header is missing and there's no basename in the URL? I answered to this question in https://github.com/marco-c/autowebcompat/pull/48#discussion_r167235321.

marco-c:
> @marco-c where to upload the downloaded extra screenshots?
and what if other's downloaded screenshots will have the same name but different files?

Which extra screenshots?

amit2rockon:
extra screenshots means ,the screenshots which are not present in current data.zip

marco-c:
> @marco-c I have made the changes here
Do you have any examples/ corner cases on which I could test the code?

With `parse_requests_response`, you don't have to parse the content-disposition header yourself. Also, you can drop the except case, `parse_requests_response` automatically implements the basename fallback.
I didn't notice that `filename_sanitized` requires an extension. Please use `filename_unsafe` then and sanitize it manually like `filename_sanitized` does (at https://github.com/g2p/rfc6266/blob/master/rfc6266.py#L121) except for the extension part.

marco-c:
> extra screenshots means ,the screenshots which are not present in current data.zip

Whenever we have new screenshots, we will upload a new data.zip file. Anyway, this seems unrelated to the current PR.

amit2rockon:
okay .

amit2rockon:
How to handle the case when the link is broken but contains baseurl ?
`parse_requests_response `  is assigning the  filename with  baseurl of the broken link also.

ManasviKundalia:
@marco-c I did the changes like you suggested replacing filename_sanitized with filename_unsafe and later sanitizing without using the extension. Please [review](https://github.com/ManasviKundalia/autowebcompat/blob/ManasviKundalia-patch-1/get_dependencies.py).

amit2rockon:
@marco-c   see these [changes  ](https://github.com/amit2rockon/autowebcompat/blob/master/get_dependencies.py)
it also solves the condition when the link is broken while it contain the baseurl.

get_dependencies.py
```diff
@@ -3,15 +3,25 @@
 import tarfile
 from zipfile import ZipFile
 import requests
+from rfc6266 import parse_requests_response
 
+def download(url):
+    response = requests.get(url, stream=True)
+    total = response.headers.get('content-length')
+    filename = parse_requests_response(response).filename_unsafe
```
marco-c: As I said in https://github.com/marco-c/autowebcompat/pull/48#issuecomment-365648565, this needs to be sanitized.

marco-c:
> How to handle the case when the link is broken but contains baseurl ?

Ignore this case, this PR is only about making the `download` function guess the name.

marco-c:
> @marco-c I did the changes like you suggested replacing filename_sanitized with filename_unsafe and later sanitizing without using the extension. Please review.

It looks good (modulo some minor issues), thanks. @ManasviKundalia anyway, since this issue is already being worked on by @amit2rockon, I suggest you pick up some other work item from the issues list.
Unfortunately only one person at a time can work on any given issue

get_dependencies.py
```diff
@@ -3,15 +3,33 @@
...
+from rfc6266 import parse_requests_response
+import posixpath
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = parse_requests_response(response).filename_unsafe
+
+    default_filename = 'file'
```
marco-c: Move the sanitizing into a separate `sanitize` function.

get_dependencies.py
```diff
@@ -3,15 +3,33 @@
...
+    if filename is None:
+        filename = default_filename
+    filename = posixpath.basename(filename)
+    filename = os.path.basename(filename)
+    filename = filename.lstrip('.')
+    if not filename:
+        filename = default_filename
+
+    if filename.rfind('.') == -1:
```
marco-c: Remove this, we don't care if the file doesn't have an extension.

get_dependencies.py
```diff
@@ -3,15 +3,33 @@
...
+        print('Couldn\'t get filename for this url')
+        return
+
     with open(filename, 'wb') as f:
-        response = requests.get(url, stream=True)
         total = response.headers.get('content-length')
 
         if total is None:
             f.write(response.content)
+            return
```
marco-c: Why are you returning here?

get_dependencies.py
```diff
@@ -22,32 +40,36 @@ def download(url, filename):
...
-with tarfile.open(name, 'r:xz') as f:
-    f.extractall('.')
-os.remove(name)
+try:
+    name = download(url)
+    print('[*] Extracting files...')
+    with tarfile.open(name, 'r:xz') as f:
+        f.extractall('.')
+    os.remove(name)
+except ValueError as e:
```
marco-c: We don't want to catch exceptions here. If there's an exception, we want the script to fail.

get_dependencies.py
```diff
@@ -22,32 +40,36 @@ def download(url, filename):
...
 
-print('[*] Extracting data.zip...')
-with ZipFile('data.zip', 'r') as z:
-    z.extractall()
+try:
+    with ZipFile('data.zip', 'r') as z:
+        print('[*] Extracting data.zip...')
+        z.extractall()
+    os.remove('data.zip')
+except ValueError as e:
```
marco-c: Same here.

requirements.txt
```diff
@@ -4,3 +4,4 @@ keras==2.1.1
 tensorflow==1.3.0
 h5py==2.7.1
 requests==2.18.4
+rfc6266==0.0.4
```
marco-c: This should go in `test-requirements.txt`, as it's only needed for development.

get_dependencies.py
```diff
@@ -3,15 +3,33 @@
...
+import posixpath
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = parse_requests_response(response).filename_unsafe
+
+    default_filename = 'file'
+    if filename is None:
```
marco-c: As I said, if we don't find the filename in the Content-Disposition header or as the basename of the URL, we want to throw an exception.

get_dependencies.py
```diff
@@ -3,11 +3,21 @@
...
+import posixpath
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = parse_requests_response(response).filename_unsafe
+    try:
+        filename = parse_requests_response(response).filename_unsafe
+    except:
```
amit2rockon: what will be the exception?

requirements.txt
```diff
@@ -4,3 +4,4 @@ keras==2.1.1
 tensorflow==1.3.0
 h5py==2.7.1
 requests==2.18.4
+
```
marco-c: Please remove this empty line.

get_dependencies.py
```diff
@@ -22,32 +32,42 @@ def download(url, filename):
...
-download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1', 'data.zip')
+download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1')
 
-print('[*] Extracting data.zip...')
 with ZipFile('data.zip', 'r') as z:
+    print('[*] Extracting data.zip...')
     z.extractall()
-
 os.remove('data.zip')
+
```
marco-c: Please remove this empty line.

get_dependencies.py
```diff
@@ -22,32 +32,42 @@ def download(url, filename):
...
 
 print('[*] Downloading data.zip...')
-download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1', 'data.zip')
+download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1')
 
-print('[*] Extracting data.zip...')
 with ZipFile('data.zip', 'r') as z:
+    print('[*] Extracting data.zip...')
     z.extractall()
-
```
marco-c: Please don't remove this line.

requirements.txt
amit2rockon: okay

get_dependencies.py
```diff
@@ -22,32 +32,42 @@ def download(url, filename):
...
     f.extractall('.')
 os.remove(name)
 
 print('[*] Downloading data.zip...')
-download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1', 'data.zip')
+download('https://www.dropbox.com/s/nkf7a6jq13gmlnu/data.zip?dl=1')
 
-print('[*] Extracting data.zip...')
 with ZipFile('data.zip', 'r') as z:
+    print('[*] Extracting data.zip...')
```
marco-c: Why did you move the print statement in here?

get_dependencies.py
```diff
@@ -22,32 +32,42 @@ def download(url, filename):
...
     url = 'https://www.dropbox.com/s/k4yifantsypy9xv/mac.tar.xz?dl=1'
-    name = 'mac.tar.xz'
 elif sys.platform.startswith('win32'):
     url = 'https://www.dropbox.com/s/xskj9rpn2fjkra8/win32.tar.xz?dl=1'
-    name = 'win32.tar.xz'
 
 print('[*] Downloading support files...')
-download(url, name)
 
+name = download(url)
```
marco-c: Move this at the same line as it was before

get_dependencies.py
```diff
@@ -3,11 +3,21 @@
...
 from zipfile import ZipFile
 import requests
+from rfc6266 import parse_requests_response
+import posixpath
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = parse_requests_response(response).filename_unsafe
```
marco-c: You're doing this twice. Do it only once without the try-except (if there is an exception, we want it to propagate to the caller).

get_dependencies.py
```diff
@@ -22,32 +32,42 @@ def download(url, filename):
                 sys.stdout.write('\r[{}{}]'.format('█' * done, '.' * (50 - done)))
                 sys.stdout.flush()
     sys.stdout.write('\n')
+    return filename
+
+
+def sanitize(filename):
+    default_filename = 'file'
```
marco-c: As I said before, we don't want a default filename. If there's no filename found, we want to throw an exception (`raise Exception('No filename could be found for this URL')`.

get_dependencies.py
amit2rockon: my code is like
    response = requests.get(url, stream=True)
    try:
        filename = parse_requests_response(response).filename_unsafe
    except Exception:
        print("Couldn\'t get file name for this URL")
         return 
what should I change here?

amit2rockon:
@marco-c  any change required?

requirements.txt
```diff
@@ -3,4 +3,4 @@ selenium==3.4.3
 keras==2.1.1
 tensorflow==1.3.0
 h5py==2.7.1
-requests==2.18.4
+requests==2.18.4
```
marco-c: Can you remove the changes from this file?

get_dependencies.py
```diff
@@ -3,11 +3,21 @@
...
 
 
-def download(url, filename):
+def download(url):
+    response = requests.get(url, stream=True)
+    filename = parse_requests_response(response).filename_unsafe
+
+    if filename is None:
+        raise Exception('No filename could be found for this URL')
+        return
```
marco-c: No need to return after raising an exception

get_dependencies.py
marco-c: You've done it 👍

requirements.txt
amit2rockon: what is changed in this line?

get_dependencies.py

amit2rockon:
@marco-c  changes has been made accordingly.

marco-c:
Thanks!

PR #50, opened 2018-02-09
Use context manager for opening and extracting tarfile

### LINKED ISSUE #49: Make get_dependencies.py use the TarFile class instead of tarfile
The TarFile class can be used as a context manager, so it is nicer.

SkullTech:
Fixes #49

### DIFF
--- a/get_dependencies.py
+++ b/get_dependencies.py
@@ -38,10 +38,9 @@ print('[*] Downloading support files...')
 download(url, name)
 
 print('[*] Extracting files...')
-f = tarfile.open(name, 'r:xz')
-f.extractall('.')
-f.close()
-
+with tarfile.open(name, 'r:xz') as f:
+    f.extractall('.')
+    f.close()
 os.remove(name)
 
 print('[*] Downloading data.zip...')
--- a/get_dependencies.py
+++ b/get_dependencies.py
@@ -40,7 +40,6 @@ download(url, name)
 print('[*] Extracting files...')
 with tarfile.open(name, 'r:xz') as f:
     f.extractall('.')
-    f.close()
 os.remove(name)
 
 print('[*] Downloading data.zip...')


get_dependencies.py
```diff
@@ -38,10 +38,9 @@ def download(url, filename):
 download(url, name)
 
 print('[*] Extracting files...')
-f = tarfile.open(name, 'r:xz')
-f.extractall('.')
-f.close()
-
+with tarfile.open(name, 'r:xz') as f:
+    f.extractall('.')
+    f.close()
```
marco-c: This is still using tarfile.open and not the TarFile class.

SkullTech: Check [this part](https://docs.python.org/3/library/tarfile.html#tarfile.TarFile) of the docs.

marco-c: This is the relevant part: https://docs.python.org/3/library/tarfile.html#tarfile-objects.

marco-c: Ah I see that `tarfile.open` can be used as a context manager too, so we can keep using it.
You don't need to close it anymore though, as it is done automatically through the context manager.

marco-c: (See https://docs.python.org/3/library/tarfile.html#tarfile.TarFile.open for reference, `tarfile.open` is a shortcut to this)

SkullTech: Oh, I see. :O The recommended way is to use `tarfile.open` though, so I think we should keep that?

marco-c: They are basically the same, as `tarfile.open` is a shortcut to `TarFile.open`. Since we are already using `tarfile.open` there's no need to change to `TarFile.open`, the only thing we need to do is make use of the context manager.

SkullTech: Okay got it.

PR #51, opened 2018-02-10
changes made for fixing the issue #47

konemshad:

PR #52, opened 2018-02-10
Update get_dependencies.py

konemshad:


marco-c:
Hello @konemshad, as I told you in #51, unfortunately the issue is already being worked on by @amit2rockon in PR #48.
You can pick up another issue if you want.

Issue #58, opened 2018-02-15
Add UTF-8 Encoding to get_dependencies.py


poush:
On running `python get_dependencies.py` we get an error as:

```
File "get_dependencies.py", line 31
SyntaxError: Non-ASCII character '\xe2' in file get_dependencies.py on line 31, but no encoding declared; see http://python.org/dev/peps/pep-0263/ for details
```

To solve this, the utf-8 encoding should be mentioned on get_dependencies.py

marco-c:
The project is requiring Python 3, where defining the encoding is not strictly necessary (it should be UTF-8 by default).

PR #59, opened 2018-02-15
Add utf-8 encoding to get_dependencies.py

poush:
Issue #58 

I was installing the project and occurred with this issue so added the fix here in PR.


marco-c:
See https://github.com/marco-c/autowebcompat/issues/58#issuecomment-366061692.

PR #62, opened 2018-02-17
Make the crawler check if screenshots for the operations were taken

### LINKED ISSUE #54: Make the crawler check if screenshots for the operations were taken
Right now, for each webcompat issue, we are just checking if we have taken screenshots for the main page. We should instead check if we have taken screenshots for all the possible operations too.

Basically, this should take care of the TODO here:
https://github.com/marco-c/autowebcompat/blob/5bb234eaac7bfbcb42bb5b6b6740d03adec6df35/collect.py#L220

Given the associated text file with the sequence of operations, we should check if a screenshot exist for every operation. When a screenshot doesn't exist, we should attempt to create it.

sviter:
Fixes #54


collect.py
```diff
@@ -211,6 +211,14 @@ def run_test(bug, browser, driver, op_sequence=None):
     return saved_sequence
 
 
+def read_sequence(bug_id):
+    elements = []
+    with open('data/%d.txt' % bug_id) as f:
+        for line in f:
```
marco-c: You can write this in a shorter way:
```Python
with open('data/%d.txt' % bug_id) as f:
    return [json.loads(line) for line in f]
```

collect.py
```diff
@@ -219,18 +227,24 @@ def run_tests(firefox_driver, chrome_driver):
...
                 with open("data/%d.txt" % bug['id'], 'w') as f:
                     for element in sequence:
                         f.write(json.dumps(element) + '\n')
+            elif not os.path.exists('data/%d_chrome.png' % bug['id']):
+                sequence = read_sequence(bug['id'])
+                run_test(bug, 'chrome', chrome_driver, sequence)
+            sequence = read_sequence(bug['id'])
+            for element_num in range(0, len(sequence)):
+                if not os.path.exists('data/%d_%d_firefox.png' % (bug['id'], element_num)):
+                    do_something(firefox_driver, sequence[element_num])
```
marco-c: Since this is a sequence of operations, if we haven't done the operation number 2, we can't just repeat 2, we have to repeat 1 first anyway.

marco-c:
I was thinking that the websites might change over time, so we can't actually rely on the sequence always being the same.
We want to retry generating the screenshots (because sometimes we implement new features of the crawler), but at the same time, if we failed to generate screenshots once, we don't want to retry too often.

So let's do it this way. Let's attempt to regenerate everything when either **a)** we haven't generated the main screenshot for Firefox or Chrome, or **b)** we haven't generated any item of the sequence for Firefox, or **c)** there are items in the Firefox sequence that we haven't generated for Chrome.

sviter:
@marco-c I changed behavior according your notes. Please check it.

collect.py
```diff
@@ -223,15 +229,21 @@ def run_tests(firefox_driver, chrome_driver):
...
-            # Assume that if we generated the main file, we also generated the one with
-            # the sequence of operations (TODO: don't assume, check!)
-            # TODO: If only Chrome is missing, don't regenerate Firefox too, but read the
-            # sequence of operations from the files.
-            if not os.path.exists('data/%d_firefox.png' % bug['id']) or\
-               not os.path.exists('data/%d_chrome.png' % bug['id']):
+            if not os.path.exists('data/%d.txt' % bug['id']) or \
+                    not os.path.exists('data/%d_firefox.png' % bug['id']) or \
+                    not os.path.exists('data/%d_chrome.png' % bug['id']):
+                sequence = run_test(bug, 'firefox', firefox_driver)
```
marco-c: This code is duplicated twice, please refactor it a bit to only write it once.

collect.py
```diff
@@ -223,15 +229,21 @@ def run_tests(firefox_driver, chrome_driver):
...
+                run_test(bug, 'chrome', chrome_driver, sequence)
+                with open("data/%d.txt" % bug['id'], 'w') as f:
+                    for element in sequence:
+                        f.write(json.dumps(element) + '\n')
+            sequence = read_sequence(bug['id'])
+            number_of_ff_scr = len(glob.glob('data/%d_*_firefox.png' % bug['id']))
+            number_of_ch_scr = len(glob.glob('data/%d_*_chrome.png' % bug['id']))
+            if len(sequence) != number_of_ff_scr or \
```
marco-c: I'm not sure this takes care of the case "we haven't generated any item of the sequence for Firefox". In this case, the sequence file is empty, so len(sequence) is 0 and number_of_ff_scr is 0 too.

sviter:
@marco-c please review changes

collect.py
```diff
@@ -218,24 +219,36 @@ def run_test(bug, browser, driver, op_sequence=None):
     return saved_sequence
 
 
+def read_sequence(bug_id):
+    with open('data/%d.txt' % bug_id) as f:
```
marco-c: Wrap this in a try except (only catch the exception thrown when the file doesn't exist). In the except case, return an empty list.

sviter:
@marco-c can you review changes? Also I think we do not achieve the main goal of this PR because we can't be sure that regenerating everything give us screenshots for full sequence.

collect.py
```diff
@@ -218,21 +219,36 @@ def run_test(bug, browser, driver, op_sequence=None):
...
+            if not os.path.exists('data/%d_firefox.png' % bug['id']) or \
+                    not os.path.exists('data/%d_chrome.png' % bug['id']) or \
+                    len(sequence) == 0 or \
+                    number_of_ff_scr != number_of_ch_scr:
+                files_to_remove = glob.glob('data/%d_*' % bug['id'])
+                for file_to_remove in files_to_remove:
+                    os.remove(file_to_remove)
                 sequence = run_test(bug, 'firefox', firefox_driver)
                 run_test(bug, 'chrome', chrome_driver, sequence)
-
```
marco-c: Nit: don't remove newline here

collect.py
```diff
@@ -218,21 +219,36 @@ def run_test(bug, browser, driver, op_sequence=None):
...
-               not os.path.exists('data/%d_chrome.png' % bug['id']):
+            # We attempt to regenerate everything when either
+            # a) we haven't generated the main screenshot for Firefox or Chrome, or
+            # b) we haven't generated any item of the sequence for Firefox, or
+            # c) there are items in the Firefox sequence that we haven't generated for Chrome.
+            sequence = read_sequence(bug['id'])
+            number_of_ff_scr = len(glob.glob('data/%d_*_firefox.png' % bug['id']))
+            number_of_ch_scr = len(glob.glob('data/%d_*_chrome.png' % bug['id']))
+            if not os.path.exists('data/%d_firefox.png' % bug['id']) or \
+                    not os.path.exists('data/%d_chrome.png' % bug['id']) or \
```
marco-c: Nit: align this and following conditions to the first `not`

collect.py
```diff
@@ -218,21 +219,36 @@ def run_test(bug, browser, driver, op_sequence=None):
...
+                    len(sequence) == 0 or \
+                    number_of_ff_scr != number_of_ch_scr:
+                files_to_remove = glob.glob('data/%d_*' % bug['id'])
+                for file_to_remove in files_to_remove:
```
marco-c: Nit: do:
```
for f in glob.iglob('data/%d_*' % bug['id'])
    os.remove(f)
```

It's clear that `f` is a file that we want to remove, so no need to be overprecise and call it `file_to_remove`.
`iglob` returns an iterator instead of returning a full list, which is better in this case.

marco-c:
> Also I think we do not achieve the main goal of this PR because we can't be sure that regenerating everything give us screenshots for full sequence.

We can't be sure, but at least we are going to try. In the worst case, we regenerate the same screenshots that we removed.
We could introduce a VERSION number for the crawler and only try to regenerate the screenshot when it changes, but it might be overly complicated for little benefit.

sviter:
@marco-c done

PR #63, opened 2018-02-17
Add unit tests for the functions in utils.py

### LINKED ISSUE #6: Add unit tests for the functions in utils.py

Shashi456:
Fixes #6

These are just very basic unit tests , which is my understanding of them. Since we are just checking each module individually and seeing if they do the job they are supposed to do. Also another thing i wanted to ask is how could i go about writing tests for functions which had no parameters, should i just check the existence of what has been returned? Thanks for your help.

@marco-c I will clean up the code and use tempfile in the upcoming commit, this was just for a basic check to see if you were expecting similar unit tests.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,4 +6,4 @@ install:
   - pip install -r test-requirements.txt
 script:
   - flake8 .
-  - pytest ./tests/test_utils.py
+  - pytest ./tests/test_*.py
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,3 +6,4 @@ install:
   - pip install -r test-requirements.txt
 script:
   - flake8 .
+  - pytest ./tests/test_utils.py
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,0 +1,42 @@
+import os
+import numpy as np
+from tempfile import TemporaryDirectory
+from autowebcompat import utils
+from PIL import Image
+
+
+def test_get_bugs():
+    bugs = utils.get_bugs()
+    assert(isinstance(bugs, list))
+
+
+def test_mkdir():
+    d = TemporaryDirectory()
+    direc_path = os.path.join(d.name + "\\test")
+    utils.mkdir(direc_path)
+    assert(os.path.isdir(direc_path))
+    utils.mkdir(direc_path)
+
+
+def test_load_image():
+    os.mkdir('data_resized')
+    fake_img = np.random.rand(30, 30, 3) * 255
+    img = Image.fromarray(fake_img.astype('uint8')).convert('RGBA')
+    img.save("./data_resized/Image.jpg")
+    img = utils.load_image("Image.jpg", "./data_resized")
+    assert(isinstance(img, np.ndarray))
+    os.remove("./data_resized/Image.jpg")
+    os.rmdir("data_resized")
+
+
+def test_read_labels():
+    labels = utils.read_labels(file_name='labels.csv')
+    assert(isinstance(labels, dict))
+
+
+def test_write_labels():
+    label = {1: 1, 2: 2}
+    d = TemporaryDirectory()
+    file_path = os.path.join(d.name + "\\test.csv")
+    utils.write_labels(label, file_name=file_path)
+    assert(os.path.exists(file_path))
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -24,6 +24,7 @@ def test_load_image(tmpdir):
     img.save(d.name + "/Image.jpg")
     img = utils.load_image("Image.jpg", d.name)
     assert(isinstance(img, np.ndarray))
+    assert(img.shape == (32, 24, 3))
 
 
 def test_read_labels():
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,8 +1,8 @@
 import os
-import numpy as np
 from tempfile import TemporaryDirectory
-from autowebcompat import utils
+import numpy as np
 from PIL import Image
+from autowebcompat import utils
 
 
 def test_get_bugs():
@@ -12,21 +12,18 @@ def test_get_bugs():
 
 def test_mkdir():
     d = TemporaryDirectory()
-    direc_path = os.path.join(d.name + "\\test")
+    direc_path = d.name + "/test"
     utils.mkdir(direc_path)
     assert(os.path.isdir(direc_path))
     utils.mkdir(direc_path)
 
 
-def test_load_image():
-    os.mkdir('data_resized')
-    fake_img = np.random.rand(30, 30, 3) * 255
-    img = Image.fromarray(fake_img.astype('uint8')).convert('RGBA')
-    img.save("./data_resized/Image.jpg")
-    img = utils.load_image("Image.jpg", "./data_resized")
+def test_load_image(tmpdir):
+    d = TemporaryDirectory()
+    img = Image.new("RGB", (30, 30))
+    img.save(d.name + "/Image.jpg")
+    img = utils.load_image("Image.jpg", d.name)
     assert(isinstance(img, np.ndarray))
-    os.remove("./data_resized/Image.jpg")
-    os.rmdir("data_resized")
 
 
 def test_read_labels():
@@ -37,6 +34,6 @@ def test_read_labels():
 def test_write_labels():
     label = {1: 1, 2: 2}
     d = TemporaryDirectory()
-    file_path = os.path.join(d.name + "\\test.csv")
+    file_path = d.name + "/test.csv"
     utils.write_labels(label, file_name=file_path)
     assert(os.path.exists(file_path))
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -70,13 +70,13 @@ def prepare_images():
 images = {}
 
 
-def load_image(fname):
+def load_image(fname, parent_dir='data_resized'):
     global images
 
     if fname in images:
         return images[fname]
 
-    img = load_img(os.path.join('data_resized', fname), target_size=(32, 24))
+    img = load_img(os.path.join(parent_dir, fname), target_size=(32, 24))
     x = img_to_array(img, data_format=keras.backend.image_data_format())
 
     images[fname] = x


tests/test_utils.py
```diff
@@ -0,0 +1,72 @@
+import os
+import csv
+import keras
+from keras.preprocessing.image import load_img, img_to_array
+
+
+def mkdir(dir_name):
```
marco-c: The functions that need to be tested should not be here.

marco-c:
>  Also another thing i wanted to ask is how could i go about writing tests for functions which had no parameters, should i just check the existence of what has been returned?

For functions that require the `data` directory to exist, don't write any test.
For other functions with no parameters (e.g. `get_bugs`), just make some basic assertion on the result.

marco-c:
> These are just very basic unit tests , which is my understanding of them. Since we are just checking each module individually and seeing if they do the job they are supposed to do.

Yes, this is what we should do.

marco-c:
> @marco-c I will clean up the code and use tempfile in the upcoming commit, this was just for a basic check to see if you were expecting similar unit tests.

OK, let me know when the PR is ready and I'll review it.

Shashi456:
@marco-c So i've written basic tests for 6 functions, just asserting the existence of the objects in case of functions with no specific parameters. For writing a couples_generator, getImageGenerator there is image_test or image_train so i have written no function for the same.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
+import os
+from autowebcompat import utils
+
+
+def test_mkdir():
+    direc_path = os.path.dirname(os.path.realpath(__file__)) + "test"
+    utils.mkdir(direc_path)
+    assert(os.path.exists(direc_path))
```
marco-c: You can assert that this is a directory.
Also, you can test calling it again to check that it doesn't fail when the directory already exists.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+label = {}
```
marco-c: Not sure you need to define this globally.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_read_labels():
+    labels = utils.read_labels(file_name='labels.csv')
+    assert('labels' in locals() or 'labels' in globals())
```
marco-c: This assertion is strange. Just assert that labels is not null and that it is a dictionary.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_write_labels():
+    utils.write_labels(label, file_name='test.csv')
+    fname = os.path.dirname(os.path.realpath(__file__)) + "\\test.csv"
+    assert(os.path.exists(fname))
```
marco-c: Also assert that you can use read_labels to read the labels back.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+images = {"Image.png": [23434, 324124]}
```
marco-c: This is unused.

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_load_image():
+    img = utils.load_image("Image.png")
+    assert(img)
```
marco-c: assert that this is a numpy 2x2 array

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_get_bugs():
+    Object = utils.get_bugs()
+    assert(Object)
```
marco-c: assert that this is a list

tests/test_utils.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def test_balance():
+    Sample = [[1, 0, 0], [0, 1, 0], [0, 0, 1]]
+    it = utils.balance(Sample)
+    assert(it)
```
marco-c: You can actually assert what balance returns here.

marco-c:
> For writing a couples_generator, getImageGenerator there is image_test or image_train so i have written no function for the same.

I'm not sure what you mean here, can you expand?

marco-c:
Nit: Can you put the tests in the same order as the functions of test_utils.py?

marco-c:
Also, don't forget to use tempfile to create the temporary directory where you store the temporary test files.

tests/test_utils.py
Shashi456: i was just asserting the existence of the variable

marco-c: You have defined it at line 15, so it clearly exists 😄

Shashi456:
@marco-c  I have added temporary directory to the unit tests wherever I am making a directory or a file. I have also made other minor changes you asked me to do. I had two doubts though, What is balance returning? and I was getting a row index out of range error while reading the written labels do you have any idea what could cause it?

tests/.pytest_cache/v/cache/lastfailed
```diff
@@ -0,0 +1,3 @@
+{
+  "test_utils.py::test_answer": true
```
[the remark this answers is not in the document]
marco-c: Indeed, and it should also be removed from the PR.

tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
+import os
+import numpy as np
+from tempfile import TemporaryDirectory
+from autowebcompat import utils
+
+
+def test_get_bugs():
+    Object = utils.get_bugs()
```
marco-c: Nit: rename `Object` to `bugs`.

tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+    assert(isinstance(Object, list))
+
+
+def test_mkdir():
+    d = TemporaryDirectory()
+    direc_path = os.path.join(d.name + "\\test")
+    utils.mkdir(direc_path)
+    assert(os.path.isdir(direc_path))
```
marco-c: Second part from https://github.com/marco-c/autowebcompat/pull/63#discussion_r168962349

tests/test_utils.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def test_load_image():
+    img = utils.load_image("Image")
```
marco-c: This file doesn't exist in the repository, so I'm not sure the test will pass.

marco-c:
> What is balance returning?

You can see the function implementation or the callers of the functions to understand. If you can't figure it out, please remove the current `test_balance` and file a follow-up issue to add a test for `balance`.

> and I was getting a row index out of range error while reading the written labels do you have any idea what could cause it?

I'm not sure, as I haven't seen what you were trying to do. If you can't figure it out, please file a follow-up issue to add the test.

marco-c:
Please also update the .travis.yml file to actually run the tests.

marco-c:
I've noticed there's a nice pytest feature that you can use instead of manually using `tempfile`: https://docs.pytest.org/en/latest/tmpdir.html.

tests/test_utils.py
Shashi456: @marco-c  earlier i had the load_image function in the file and i was using a random dictionary to just check if the code was working, Because an empty file name would result in a parameter not declared error. Should i remove the test ?

tests/test_utils.py
Shashi456: @marco-c  Instead of writing a test for it , would it not make more sense to change the mkdir function to check if a directory with the same name exists and not make it if it does? since the function the doesnt return any error message it is pretty hard to assert on anything . If some error message were returned i could assert on the same .

marco-c: You don't need to add an assertion, the assertion is implied in the fact that `utils.mkdir` doesn't throw an exception.

tests/test_utils.py
marco-c: You can add a `parent_dir` argument to `load_image`, with the default being `data_resized`.
Then in the test you can create a fake image in the temporary directory, pass the temporary directory as parent_dir to load_image and then assert that the image is loaded as a 2x2 numpy array with the correct dimensions.

Shashi456:
@marco-c  i've made all the minor changes you asked me to, i have removed the balance test and will open a pr for the same, because balance seems to be interacting with Couples-generator but we aren't writing tests for functions which need the data directory. One last thing is the error in the Travis-cl build it is giving an image not found error with the Load_image test, while I am able to pass the test successfully on my pc. Any Thoughts?

marco-c:
> @marco-c i've made all the minor changes you asked me to

It seems you haven't done https://github.com/marco-c/autowebcompat/pull/63#issuecomment-366649639 and https://github.com/marco-c/autowebcompat/pull/63#discussion_r169072493.

> i have removed the balance test and will open a pr for the same, because balance seems to be interacting with Couples-generator but we aren't writing tests for functions which need the data directory.

The balance function is generic, it doesn't require CouplesGenerator to work. Anyway yes, please open an issue to add a test for this function and an issue for each function in utils.py that isn't tested yet.

> One last thing is the error in the Travis-cl build it is giving an image not found error with the Load_image test, while I am able to pass the test successfully on my pc. 

Doing https://github.com/marco-c/autowebcompat/pull/63#discussion_r169072493 will fix the problem.

Shashi456:
@marco-c i did modify both the load image file with the parent directory and try it with both temporary and an actual directory.  I will look further into, thankyou.

marco-c:
> @marco-c i did modify both the load image file with the parent directory and try it with both temporary and an actual directory. I will look further into, thankyou.

Also don't download the image from the network, it's not great for unit tests (what if the image is removed from the website? The tests will start failing). Create a fake image using PIL.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    utils.mkdir(direc_path)
+    assert(os.path.isdir(direc_path))
+    utils.mkdir(direc_path)
+
+
+def test_load_image():
+    os.mkdir('data_resized')
+    fake_img = np.random.rand(30, 30, 3) * 255
+    img = Image.fromarray(fake_img.astype('uint8')).convert('RGBA')
+    img.save(".\data_resized\Image.jpg", "JPEG")
```
marco-c: Don't specify the format, it's automatically detected from the file extension.

Shashi456:
@marco-c  I'll open the balance PR , as soon as this is merged. Everything works and i've done all changes needed. Thankyou

Shashi456:
@marco-c i will open the balance PR as soon as this is merged, i have done all changes  you've asked me for . Thankyou

marco-c:
> @marco-c i will open the balance PR as soon as this is merged, i have done all changes you've asked me for . Thankyou

I still don't see https://github.com/marco-c/autowebcompat/pull/63#issuecomment-366649639 :smile:

Also, in `test_load_image` you should use a temporary directory and not manually create and delete `data_resized`, which would mess the working directory of anyone which runs the tests.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    direc_path = os.path.join(d.name + "\\test")
+    utils.mkdir(direc_path)
+    assert(os.path.isdir(direc_path))
+    utils.mkdir(direc_path)
+
+
+def test_load_image():
+    os.mkdir('data_resized')
+    fake_img = np.random.rand(30, 30, 3) * 255
+    img = Image.fromarray(fake_img.astype('uint8')).convert('RGBA')
```
marco-c: Use `Image.new` instead of creating a numpy array and then convering it into an image.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    labels = utils.read_labels(file_name='labels.csv')
+    assert(isinstance(labels, dict))
+
+
+def test_write_labels():
+    label = {1: 1, 2: 2}
+    d = TemporaryDirectory()
+    file_path = os.path.join(d.name + "\\test.csv")
+    utils.write_labels(label, file_name=file_path)
+    assert(os.path.exists(file_path))
```
marco-c: Remember to file an issue about adding a test that writes labels and then reads them back.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_get_bugs():
+    bugs = utils.get_bugs()
+    assert(isinstance(bugs, list))
+
+
+def test_mkdir():
+    d = TemporaryDirectory()
+    direc_path = os.path.join(d.name + "\\test")
```
marco-c: You're calling `os.path.join` with a single string, which is basically useless.
Either do `os.path.join(d.name, 'test')` or `d.name + '\test'`.

tests/test_utils.py
```diff
@@ -0,0 +1,42 @@
+import os
+import numpy as np
+from tempfile import TemporaryDirectory
+from autowebcompat import utils
+from PIL import Image
```
marco-c: Sort the imports in this way:
1) Built-in modules;
2) Modules from third-party libraries;
3) Our modules.

Shashi456:
@marco-c do you necessarily want the new pytest feature and not my hack around it ?

.travis.yml
```diff
@@ -6,3 +6,4 @@ install:
   - pip install -r test-requirements.txt
 script:
   - flake8 .
+  - pytest ./tests/test_utils.py
```
marco-c: This is going to execute `test_utils.py` only, please make it run any test that is in the `tests` directory, so we don't have to update this line the next time we add a test.

marco-c:
> @marco-c do you necessarily want the new pytest feature and not the Temporary Directory () way?

The pytest feature is nicer, so I would prefer it. If you can't, please file an issue for someone else to do it.

Shashi456:
Would you want the PR's to be referenced to this?

marco-c:
> Would you want the PR's to be referenced to this? @marco-c

Do you mean the issues? Yes

tests/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+def test_load_image(tmpdir):
+    d = TemporaryDirectory()
+    img = Image.new("RGB", (30, 30))
+    img.save(d.name + "/Image.jpg")
+    img = utils.load_image("Image.jpg", d.name)
+    assert(isinstance(img, np.ndarray))
+
+
+def test_read_labels():
+    labels = utils.read_labels(file_name='labels.csv')
```
marco-c: Create a fake labels.csv file in a temporary directory, so you can assert more things about the result.

marco-c: Actually, don't do it, we'll do it as part of the other test that writes and reads labels.

tests/test_utils.py
```diff
@@ -0,0 +1,39 @@
...
+    assert(os.path.isdir(direc_path))
+    utils.mkdir(direc_path)
+
+
+def test_load_image(tmpdir):
+    d = TemporaryDirectory()
+    img = Image.new("RGB", (30, 30))
+    img.save(d.name + "/Image.jpg")
+    img = utils.load_image("Image.jpg", d.name)
+    assert(isinstance(img, np.ndarray))
```
marco-c: You can assert the dimensions of the numpy array too.

.travis.yml
marco-c: Don't forget about this, @Shashi456

PR #64, opened 2018-02-17
Project restructred for better working with modules

### LINKED ISSUE #37: Rearrange the project directory structure
Rearrange it into a structure like [this](https://docs.pytest.org/en/latest/goodpractices.html#tests-outside-application-code), as suggested by @marco-c [here](https://github.com/marco-c/autowebcompat/pull/36#discussion_r166479187). This will facilitate testing, among other things.

poush:
Fixes #37 

The scripts are left in the root to not to add the code to import from parallel directory inside the scripts subdirectory. This PR mainly transfers the network.py and utils.py to the `autowebcompat` directory and updates the files importing them.

Let me know where we can improve it more.


setup.cfg
```diff
@@ -1,6 +1,6 @@
 [flake8]
 # no need to traverse these directories
-exclude = .git,__pycache__,data,data_resized,tools
+exclude = .git,__pycache__,data,data_resized,tools,ENV
```
marco-c: What is ENV?

.gitignore
```diff
@@ -9,4 +9,4 @@ tools/geckodriver
 tools/chromedriver
 
 geckodriver.log
-.DS_Store
+.DS_Store
```
marco-c: Can you remove the changes to this file from the PR?

setup.cfg
poush: Sorry, At first time I created my virtualenv directory (/ENV) inside the project so had to add it for running flake. I missed it to remove.

.gitignore
poush: Yeah sure.

PR #85, opened 2018-02-21
Don't reference `child_attributes` before assignment

sagarvijaygupta:
If no ```children``` are found in body this results in ```UnboundLocalError: local variable 'child_attributes' referenced before assignment```

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -111,7 +111,7 @@ def do_something(driver, elem_attributes=None):
 
         for child in children:
             # Get all the attributes of the child.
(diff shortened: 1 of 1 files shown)


collect.py
```diff
@@ -97,6 +97,7 @@ def get_all_attributes(driver, child):
 
 def do_something(driver, elem_attributes=None):
     elem = None
+    child_attributes = None
```
marco-c: Don't add this, but make the function return `elem_attributes` and rename `child_attributes` to `elem_attributes` at line 114.

PR #88, opened 2018-02-22
Choose all possible elements to interact with.

### LINKED ISSUE #45: Instead of randomly choosing an element to interact with, choose all possible elements
Currently, we are choosing elements to interact with in a random fashion.
Instead, we should explore all possible paths (up to a maximum level that we need to decide).

sagarvijaygupta:
Fixes #45.


sagarvijaygupta:
Line numbering for images has `0 based indexing`
I have renamed the present `WEBCOMPATID_SEQUENCE.png` according to new convention of `WEBCOMPATID_LINENUMBER.png` where `line number` is the tree to be followed for the screenshot in its `WEBCOMPAT.txt` file as suggested in #45 using the 3rd way.

sagarvijaygupta:
@marco-c please review this!

sagarvijaygupta:
[6817.zip](https://github.com/marco-c/autowebcompat/files/1754453/6817.zip)
@marco-c  These are the screenshots with file.txt for bug_id 6817

sagarvijaygupta:
@marco-c please review it?

marco-c:
Hey @sagarvijaygupta, since this is a larger change, it requires more time to be reviewed (and I haven't been able to allocate enough time for reviewing it in the past three days).
I will try to get to it tomorrow or Tuesday.
I'm also thinking about improving the way we store the data, so that we can also see the history of how the naming changes.

sagarvijaygupta:
@marco-c Oh okay! I thought there was some issue in the PR because some checks were not successful. Thanks for the reply! :smile:

collect.py
```diff
@@ -112,13 +113,23 @@ def do_something(driver, elem_attributes=None):
         for child in children:
             # Get all the attributes of the child.
             elem_attributes = get_all_attributes(driver, child)
+            path_to_follow.append(elem_attributes)
+
+            # We check the nodes if previously visited or not
+            if frozenset([frozenset(attributes) for attributes in path_to_follow]) in visited_path:
```
marco-c: I think this might be imprecise. What if there are two buttons with the same attributes on two different pages? We'll consider them as being the same button.
I would remove the visited checks until we can think of a more precise solution.

There's also a problem with the current code. `attributes` is an object where the keys are the names of the attributes and the values are the values of the attributes. frozenset({'key': 'value'}) is frozenset({'key'}). We should consider both the names and the values to identify elements and not only the names.

The code is also not very readable. I suggest moving the visited check and the visited addition into separate functions (something like `was_visited` and `visit`).

collect.py
```diff
@@ -180,50 +191,77 @@ def screenshot(driver, file_path):
...
+    print('Testing %s (bug %d) in %s' % (bug['url'], bug['id'], "firefox"))
 
     try:
-        driver.get(bug['url'])
+        firefox_driver.get(bug['url'])
     except TimeoutException as e:
         # Ignore timeouts, as they are too frequent.
         traceback.print_exc()
         print('Continuing...')
+        return
```
marco-c: We shouldn't return in case of timeouts, they are extremely common.

collect.py
marco-c: Oh, I see. You were not trying to implement a way to check whether we already visited a page, but whether we already followed a path.
OK, makes more sense then. The latest two paragraphs still apply.

collect.py
```diff
@@ -180,50 +191,77 @@ def screenshot(driver, file_path):
     image.save(file_path)
 
 
-def run_test(bug, browser, driver, op_sequence=None):
-    print('Testing %s (bug %d) in %s' % (bug['url'], bug['id'], browser))
+def lines_written_in_file(bug_id):
```
marco-c: Rename this function to count_sequence or count_lines.

marco-c:
To clarify the variables, I would rename `visited_path` to `visited_paths` and `path_to_follow` to `current_path`.

I think we shouldn't use a frozenset, which doesn't have an order.

`visited_paths` should be a set of lists of nodes. `current_path` should be a list of nodes. A node is an object containing the attributes.

To add a note to the current path, just do `current_path.append(attributes)`.
To add a path to the visited_paths, just do `visited_paths.add(current_path)`.
To check if a path is in the visited_paths, just do `current_path in visited_paths`.

collect.py
marco-c: There was another comment from me here, which somehow got removed.
In the comment, I was saying I thought you were trying to do something else and so you could discard the first paragraph of my first comment.

sagarvijaygupta:
[6822.txt](https://github.com/marco-c/autowebcompat/files/1767317/6822.txt)
@marco-c  This is the text file generated for issue 6822 after the required changes. It takes some time for ```jump_back``` because we start from the beginning. Is it okay? Or can we think of something better.

sagarvijaygupta:
@marco-c Please review it!

marco-c:
When you want to request me to re-review, click on "Dismiss review", so I know which PRs I have to review from the list.

sagarvijaygupta:
@marco-c I don't see the "Dismiss review" link. Only "See review" is visible.

sagarvijaygupta:
@marco-c please review it!

marco-c:
>  @marco-c This is the text file generated for issue 6822 after the required changes. It takes some time for jump_back because we start from the beginning. Is it okay? Or can we think of something better.

Sorry if I haven't reviewed this yet, but the slowdown worries me a bit. I suggest you work on something else and we keep this open for a while until we decide what to do.

sagarvijaygupta:
@marco-c thanks for the reply! I will work on some other issue.
For this, I was thinking to interact with elements every time in new tab. This will increase the speed as we don't have to regenerate the whole path but might not be able to work with all intractable elements like text and email.

sagarvijaygupta:
@marco-c There is another idea which will remove the overhead of traversing from start. If possible I will try to create and independent copy of driver instance and work on it separately.

marco-c:
> @marco-c There is another idea which will remove the overhead of traversing from start. If possible I will try to create an independent copy of driver instance and work on it separately.

I think it won't be possible, but I would be happy if I'm wrong :)

sagarvijaygupta:
@marco-c Just for the record I implemented it. You are right, its not possible that way. :frowning_face:

sagarvijaygupta:
@marco-c I have another solution. I will implement it after your suggestion.
Main overhead that we encounter is in this:
```
def jump_back(current_path, firefox_driver, chrome_driver, visited_paths, bug):
    firefox_driver.get(bug['url'])
    chrome_driver.get(bug['url'])
    for elem_attributes in current_path:
        do_something(firefox_driver, visited_paths, current_path, elem_attributes)
        do_something(chrome_driver, visited_paths, current_path, elem_attributes)
```
When we call this for elements without element id's in `do_something` this part is executed which takes  good amount of time which I found while trying to implement the previous suggestion.

            buttons = body.find_elements_by_tag_name('button')
            selects = body.find_elements_by_tag_name('select')
            children = buttons + links + inputs + selects
If in `jump_back` we use `xpath` (different for different browsers which will saved along with `elem_attributes` in `current_path`) the above overhead will be removed and we will just find the element by xpath. Without finding all elements and iterating them over again to match with `elem_attributes`.

In most places on internet it is written that finding element by xpath is slower (and not preferable for testing but since our DOM won't change we can use it), but in my search I _didn't find a place where they have specified absolute xpath is slower_. _Relative xpath_ in general or with parameters like _contains_ or _startswith_ etc should be slower but I am not sure whether absolute xpath also has to be slower? Should I try and go on with it? Overall it will increase the speed since we won't be iterating over all children to match attributes but I wanted to know that should I use xpath only, or css selector or something else?

marco-c:
It seems reasonable to me, I would expect using the XPath to be faster than what we're doing right now, since we're basically reimplementing XPath from outside the browser and in a "stupid" way (just looking at all elements).

The only problem that could arise is that there could be cases where with the xpath we don't find an element that we would have found with the current approach (which could happen if an xpath is valid in one browser and not in the other). A possible solution would be to use XPath first and fallback on the old technique if XPath fails to find the element.

Anyway, I'd start with the simple approach (using the current slow technique) and then in another PR evaluate switching to use XPath.

sagarvijaygupta:
> The only problem that could arise is that there could be cases where with the xpath we don't find an element that we would have found with the current approach (which could happen if an xpath is valid in one browser and not in the other).

I didn't get it. 
What I am saying is that we follow the current "stupid" approach only but with some modification. We choose new element to interact with on the basis of "stupid" approach and save its xpath for the both browsers (even if they are different it doesn't matter because we will use the xpath saved for that browser only). Now when we **jump_back** ie start from getting the URL and reach the current level, we find elements by the saved xpaths. Are we on the same page?

> Anyway, I'd start with the simple approach (using the current slow technique) and then in another PR evaluate switching to use XPath.

This PR is ready with simple approach.

marco-c:
> I didn't get it.
> What I am saying is that we follow the current "stupid" approach only but with some modification. We choose new element to interact with on the basis of "stupid" approach and save its xpath for the both browsers (even if they are different it doesn't matter because we will use the xpath saved for that browser only). Now when we jump_back ie start from getting the URL and reach the current level, we find elements by the saved xpaths. Are we on the same page?

Yes, I'm talking about the jump_back too. The only problem I see is that if you collect the XPath on Firefox and then try to apply it on Chrome, it might not work.

sagarvijaygupta:
@marco-c I think there is some confusion or I am not understanding some part.
What I am saying is that:

1. Since we have the element for which we got the `elem_attributes` I get xpath in firefox and save it.
2. I follow the "stupid" approach only, for chrome and get the same element for `elem_attributes`as I am getting for firefox and for that element get another xpath in chrome.
3. We have two paths: 
 a. chrome_xpaths = [list of xpaths required to get to present level in chrome]
 b. firefox_xpaths = [list of xpaths required to get to present level in firefox]
3. Now when we call jump_back instead of using the "stupid" approach to traverse we use stored xpaths differently for different browsers.

> The only problem I see is that if you collect the XPath on Firefox and then try to apply it on Chrome, it might not work.
I won't be applying the xpath I got in Firefox to chrome, I will be using chrome's xpath for it

marco-c:
Yes, if we use different XPaths then we should be OK.

marco-c:
The problem I was talking about only applies if we are using the same xpath for both Firefox and Chrome.

collect.py
```diff
@@ -231,49 +256,72 @@ def get_screenshot_and_domtree(driver, bug_id, browser, seq_no=None):
...
         print('Continuing...')
 
-    return saved_sequence
+    get_screenshot_and_domtree(chrome_driver, str(bug['id']), 'chrome')
 
+    visited_paths = []
+    current_path = []
+    while True:
+        elem_properties = do_something(firefox_driver, visited_paths, current_path)
+        print(elem_properties)
```
marco-c: Print as before, to have more verbose output: `print('  - Using %s' % elem_properties)`

sagarvijaygupta:
@marco-c By this PR images we have presently are not renamed according to the convention in this PR.

marco-c:
> @marco-c By this PR images we have presently are not renamed according to the convention in this PR.

Why? What changed in the convention?

marco-c:
> @marco-c as it was discussed in #45 we decided the new naming convention on the basis of linenumber in its corresponding txt file. So both txt file and image names are not according to what we proposed.
We preferred the third method you suggested.

The old names should still be compatible though, no?

With this PR, the name is composed by:
`get_screenshot_and_domtree(firefox_driver, str(bug['id']), 'firefox', str(line_number))`
while previously it was:
`get_screenshot_and_domtree(driver, str(bug['id']), browser, str(i))`

sagarvijaygupta:
No actually.
Previously              ->      Presently
>bug_0_browser.png -> bug_0_browser.png
bug_1_browser.png -> bug_2_browser.png
bug_2_browser.png -> bug_5_browser.png

And also the text file
Previously
>elem_attr1
elem_attr2

Presently
>elem_attr1
>
>elem_attr1
elem_attr2

marco-c:
Right, we'll need to write a script to migrate the data then.

marco-c:
Actually @sagarvijaygupta, the names *will* be different, but aren't we still able to read them? This is what I meant by `compatible`.

sagarvijaygupta:
I will make the required change. But I wanted to know how to push images and txt files to your gitlab. I tried to push some images with full_page_screenshots but wasn't able to.

marco-c:
> I will make the required change. But I wanted to know how to push images and txt files to your gitlab. I tried to push some images with full_page_screenshots but wasn't able to.

What was the problem?

sagarvijaygupta:
>Actually @sagarvijaygupta, the names will be different, but aren't we still able to read them? This is what I meant by compatible.
You are right we will be able to read them but there will be inconsistency which might result in some problems afterwards like
>bug_1_browser.png -> bug_2_browser.png
bug_2_browser.png -> bug_5_browser.png

`bug_2_browser.png` will mean different if created previously and different if created now.

marco-c:
Yes, the meaning of the individual items would be different, but we are not losing information that we require right now.
Doing the migration is useful because we might need the info in the future, but it doesn't break anything right now (so people can keep labeling, training, etc. on on current master).

sagarvijaygupta:
Yeah! It is good to go in that way. We can leave present images and text files as it is. It will be better this way as you suggested since we won't require to download data again (or run the script). 
So we will skip the migration as of now?

marco-c:
Yes, we can skip it for the time being :)
Maybe start to write the script, but we won't actually apply it yet.

Issue #81, opened 2018-02-20
Add support for input type 'color'


poush:
Like  different actions we can add `color` input type as well.

poush:
@marco-c Are you sure that 'testing' label is relevant for this PR?

marco-c:
> @marco-c Are you sure that 'testing' label is relevant for this PR? I think 'crawler' is good one here.

Good catch!

Trion129:
Hi can I take this issue? would help me get started with the code :) Planning to contribute to autowebcompat in GSoC

marco-c:
> Hi can I take this issue? would help me get started with the code :) Planning to contribute to autowebcompat in GSoC

Sure, go ahead!

Trion129:
Can be closed, as it is fixed

PR #91, opened 2018-02-24
Added color support and changing code

Trion129:
Added "color" input type support in crawler to solve #81 

First the crawler clicks on the input, this puts it in view and then uses the included JS code to change the input value to red.

Verified on http://www.htmlquick.com/reference/tags/input-color.html

## Firefox screenshot:
image

## Chrome Screenshot
image

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -165,7 +165,6 @@ def do_something(driver, elem_attributes=None):
             elem.clear()
             elem.send_keys('quick search')
         elif input_type == 'color':
-            driver.execute_script("arguments[0].scrollIntoView();", elem)
             driver.execute_script("arguments[0].value = '#ff0000'", elem)
         else:
             raise Exception('Unsupported input type: %s' % input_type)
--- a/collect.py
+++ b/collect.py
@@ -164,6 +164,9 @@ def do_something(driver, elem_attributes=None):
         elif input_type == 'search':
             elem.clear()
             elem.send_keys('quick search')
+        elif input_type == 'color':
+            elem.click()
+            driver.execute_script("arguments[0].setAttribute('value', '#FF0000')", elem)
         else:
             raise Exception('Unsupported input type: %s' % input_type)
 
--- a/collect.py
+++ b/collect.py
@@ -165,8 +165,8 @@ def do_something(driver, elem_attributes=None):
             elem.clear()
             elem.send_keys('quick search')
         elif input_type == 'color':
-            elem.click()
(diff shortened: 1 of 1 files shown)


collect.py
```diff
@@ -164,6 +164,9 @@ def do_something(driver, elem_attributes=None):
         elif input_type == 'search':
             elem.clear()
             elem.send_keys('quick search')
+        elif input_type == 'color':
+            elem.click()
```
marco-c: Clicking on the element shouldn't be needed.

collect.py
```diff
@@ -164,6 +164,9 @@ def do_something(driver, elem_attributes=None):
...
+            driver.execute_script("arguments[0].setAttribute('value', '#FF0000')", elem)
```
marco-c: You can use `arguments[0].value = '#FF0000'`.
Or even better, use the Selenium API to do it

Trion129: I tried to find a set_attribute function in Selenium WebDriver and WebElement API but it wasn't present. So I went for javascript's setAttribute.
The `.value` notation looks cleaner to me, will try it :)
I'll replace the `.click()` with an ActionChain, so I move to the element and set the value.

marco-c: If you use `.value`, you don't need to do anything else before

Trion129:
Tried various approaches:

First approach:
1. click()
2. setAttribute using executescript
Problem: There was inconsistency in the screenshot on the position of input, might have caused problem in model training. Also click isn't a very clean procedure for this purpose.

Second approach:
1. Directly setting `.value`
Problem: Unfortunately, Doesn't scroll the element in view

Third approach:
1. Use ActionChain to move to the element
2. Setting `.value`
Problem: For the example website, it errored out giving out of viewport error, weird!

Fourth approach:
1. Used JS to call `.scrollIntoView()`
2. Set the `.value`
Screenshots are consistent and show the color input as expected! More beneficial for model to learn from.

## Chrome:
image

## Firefox:
image

marco-c:
The scrollIntoView is a good idea! It should be implemented in a separate PR though, and for all kinds of elements (not just color).

collect.py
```diff
@@ -164,6 +164,9 @@ def do_something(driver, elem_attributes=None):
         elif input_type == 'search':
             elem.clear()
             elem.send_keys('quick search')
+        elif input_type == 'color':
+            driver.execute_script("arguments[0].scrollIntoView();", elem)
```
marco-c: Let's remove this, we will add it in a separate PR.

PR #93, opened 2018-02-27
Add some unit tests for network.py

### LINKED ISSUE #76: Add tests for network.py

vibss2397:
Hey, 
as discussed, tried to make the test_network file.

In addition, I was going through the code base and wanted to ask in the pretrain file 
``def couples_generator(images):
    # for image_couple in itertools.combinations_with_replacement(images, 2):
    for image_couple in itertools.combinations(images, 2):
        yield image_couple, 1 if are_same_site(image_couple[0], image_couple[1]) else 0 ``

Isn't this doing the opposite of what was discussed in the paper from Hadsell-et-al.'06 wherein more similar images were closer to each other

### DIFF
--- a/tests/test_network.py
+++ b/tests/test_network.py
@@ -28,5 +28,6 @@ def test_eucl_distance_output_shape():
 def test_contrastive_loss():
     euclid_dist=calc_dist(a,b)
     loss=network.contrastive_loss(euclid_dist,1)
-    assert(loss>0)
+    eval=K.eval(loss)
+    assert(eval>0)
 


tests/test_network.py
```diff
@@ -0,0 +1,33 @@
...
+    dist=np.sum(np.square(a-b))
+    dist_sqroot=np.sqrt(max(dist,K.epsilon()))
+    return dist_sqroot
+
+def test_eucledian_distance():
+    image0 = K.variable(value=a)
+    image1 = K.variable(value=b)
+    dist = network.euclidean_distance([image0,image1])
+    evaluate=K.eval(dist)
+    assert(evaluate==calc_dist(a,b))
```
marco-c: Don't use calc_dist, use hardcoded values.
Also, with floats you shouldn't check for exact equality.

tests/test_network.py
```diff
@@ -0,0 +1,33 @@
...
+    vect=[a.shape,b.shape]
+    shape=network.eucl_dist_output_shape(vect)
+    assert(shape==(1,1))
+
+
+def test_contrastive_loss():
+    euclid_dist=calc_dist(a,b)
+    loss=network.contrastive_loss(euclid_dist,1)
+    eval=K.eval(loss)
+    assert(eval>0)
```
marco-c: We should check the value like for the euclidean distance.

marco-c:
> In addition, I was going through the code base and wanted to ask in the pretrain file
def couples_generator(images): # for image_couple in itertools.combinations_with_replacement(images, 2): for image_couple in itertools.combinations(images, 2): yield image_couple, 1 if are_same_site(image_couple[0], image_couple[1]) else 0

> Isn't this doing the opposite of what was discussed in the paper from Hadsell-et-al.'06 wherein more similar images were closer to each other

pretrain.py is trying to solve the problem of identifying screenshots that belong to the same website vs screenshots that don't belong to the same website. That snippet of code is generating the screenshot couples with their labels.
Why is this the opposite of what was discussed in the paper?

P.S.: This isn't related to the current PR. Please file an issue when you have an observation, otherwise I might not notice it.

vibss2397:
@marco-c  
while uploading the files there was an error so I deleted the old one and forked it again, and now it is showing unknown, any help on what to do now would really be helpful. Since the build was failing I started acting panicky and forked the repository again. I'm really sorry for that, it won't happen again. How do i go on about this

marco-c:
If you are having problems updating the PR, feel free to close this one and open a new one.

PR #95, opened 2018-02-28
Add some unit tests for network.py

vibss2397:
Created tests for some of the functions in network,py, do check and tell if any changes are required.

(Earlier discussion about this PR is in #93)


vibss2397:
@marco-c i wanted to ask are there some other functions in this file which need to be tested?
Also, my uni exams are coming up so was a bit busy for the past few days

marco-c:
> @marco-c i wanted to ask are there some other functions in this file which need to be tested?

Yes, we should try to test as many functions as possible. This PR is only adding tests for a subset of the functions.
You can do it in a follow-up PR.

tests/test_network.py
```diff
@@ -0,0 +1,44 @@
...
+arr1 = np.array([[1, 1]], dtype=np.float32)
+arr2 = np.array([[0, 1]], dtype=np.float32)
+
+arr3 = np.array([[0, 0]], dtype=np.float32)
+arr4 = np.array([[3, 2]], dtype=np.float32)
+
+
+def test_eucledian_distance():
+    dist = network.euclidean_distance([arr1, arr2])
+    evaluate = K.eval(dist)
```
marco-c: Remove the `evaluate` variable and just do the assert here

tests/test_network.py
```diff
@@ -0,0 +1,44 @@
...
+    dist2 = network.euclidean_distance([arr3, arr4])
+    evaluate2 = K.eval(dist2)
```
marco-c: Same here

tests/test_network.py
```diff
@@ -0,0 +1,44 @@
...
+    assert (evaluate == 1)
+    assert (math.floor(evaluate2 * 100) == 360)
+
+
+def test_eucl_distance_output_shape():
+    vect = [arr1.shape, arr2.shape]
+    shape = network.eucl_dist_output_shape(vect)
+    assert (shape == (1, 1))
```
marco-c: Don't create `vect` and `shape` variables, just assert the call directly like this:
```Python
assert(network.eucl_dist_output_shape([arr1.shape, arr2.shape]) == (1, 1))
```

Also test arr4

tests/test_network.py
```diff
@@ -0,0 +1,44 @@
...
+
+
+def test_contrastive_loss():
```
marco-c: This seems overly complex. As a unit test, this should just test the `contrastive_loss` function.

vibss2397:
@marco-c for the models in the network file, do we test the shape of inputs, outputs or they have weights or what do we want to test?

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+    assert (math.floor(K.eval(dist2) * 100) == 360)
+
+
+def test_eucl_distance_output_shape():
+    vect = [arr1.shape, arr2.shape]
+    assert (network.eucl_dist_output_shape(vect) == (1, 1))
+
+
+def test_contrastive_loss():
+    label = [0, 1]  # possible values for output
```
marco-c: Why defining an array here? Just use the raw values when needed.

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+    euclid_dist = network.euclidean_distance([arr1, arr2])  # it returns 1
```
marco-c: You're testing `contrastive_loss` here, so there's no pointing in calling `euclidean_distance` if you already know its result.

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+
+def test_eucledian_distance():
+    dist = network.euclidean_distance([arr1, arr2])
+    dist2 = network.euclidean_distance([arr3, arr4])
+    assert (K.eval(dist) == 1)
+    assert (math.floor(K.eval(dist2) * 100) == 360)
+
+
+def test_eucl_distance_output_shape():
+    vect = [arr1.shape, arr2.shape]
```
marco-c: Already told in my earlier comment, please remove the `vect` variable.

marco-c: Also, as I said before, also test arr4

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+arr2 = np.array([[0, 1]], dtype=np.float32)
+
+arr3 = np.array([[0, 0]], dtype=np.float32)
+arr4 = np.array([[3, 2]], dtype=np.float32)
+
+
+def test_eucledian_distance():
+    dist = network.euclidean_distance([arr1, arr2])
+    dist2 = network.euclidean_distance([arr3, arr4])
+    assert (K.eval(dist) == 1)
```
marco-c: Nit: this is a weird way to arrange the assertions. Put the assertion related to a `euclidean_distance` call right after the call itself.

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+arr1 = np.array([[1, 1]], dtype=np.float32)
+arr2 = np.array([[0, 1]], dtype=np.float32)
+
+arr3 = np.array([[0, 0]], dtype=np.float32)
+arr4 = np.array([[3, 2]], dtype=np.float32)
+
+
+def test_eucledian_distance():
+    dist = network.euclidean_distance([arr1, arr2])
+    dist2 = network.euclidean_distance([arr3, arr4])
```
marco-c: Use `dist` and not `dist2` when you move the assertions

tests/test_network.py
```diff
@@ -0,0 +1,37 @@
...
+    loss1 = network.contrastive_loss(label[0], euclid_dist)
+    loss2 = network.contrastive_loss(label[1], euclid_dist)
+    assert (K.eval(loss1) == 0)  # since the output and the label are far away
+    assert (K.eval(loss2) == 1)  # since the output and the label are close
+
+
+def test_accuracy():
+    y_pred = np.array([0, 0.4, 0.7, 0.6, 0.5])
+    y_true = np.array([0, 1, 0, 0, 1])
+    assert (math.floor(K.eval(network.accuracy(y_true, y_pred)) * 100) == 60)
```
marco-c: Nit: no space between assert and `(`, here and in the rest of the file.

vibss2397:
@marco-c done with the changes

tests/test_network.py
```diff
@@ -0,0 +1,33 @@
...
+
+
+def test_eucledian_distance():
+    dist = network.euclidean_distance([arr1, arr2])
+    assert (K.eval(dist) == 1)
+
+
+def test_eucl_distance_output_shape():
+    assert(network.eucl_dist_output_shape([arr1.shape, arr2.shape]) == (1, 1))
+    assert(network.eucl_dist_output_shape([arr3.shape, arr4.shape]) == (1, 1))
```
marco-c: Can you add a test with another array that has a different shape? Just to have a different output than just (1, 1).

PR #100, opened 2018-03-02
Added support for select tags

vikasmahato:


vikasmahato:
I can see that collect.py gets the urls to be crawled from webcompatdata-bzlike.json and the screenshots are stored in data/ 
So should I add an entry to webcompatdata-bzlike.json for a URL that contains a select tag? Or is there a better approach?
I intend to test this url https://vikasmahato.github.io/elements.html

marco-c:
You can just temporarily modify collect.py to use the URL you want.

vikasmahato:
Chrome Screenshot
9_2_chrome

Firefox Screenshot
9_2_firefox

collect.py
```diff
@@ -105,7 +105,8 @@ def do_something(driver, elem_attributes=None):
         buttons = body.find_elements_by_tag_name('button')
         links = body.find_elements_by_tag_name('a')
         inputs = body.find_elements_by_tag_name('input')
-        children = buttons + links + inputs
+        selects = body.find_elements_by_tag_name('select')
```
marco-c: Why not looking for the `option` elements directly?

collect.py
```diff
@@ -170,6 +172,12 @@ def do_something(driver, elem_attributes=None):
             driver.execute_script("arguments[0].value = '#ff0000'", elem)
         else:
             raise Exception('Unsupported input type: %s' % input_type)
+    elif elem.tag_name == 'select':
+        elem.click()
```
marco-c: Do we need to click on the `select` element?

vikasmahato: I thought that we are trying to emulate user interaction with the webpage, hence I clicked the select element to display the select options. This is also the reason that I am not looking for 'option' elements directly. Please correct me if I am wrong.

marco-c: If it is necessary to click on the select element first before clicking on the option element, then it makes sense to do that.
If it isn't necessary, then there's no need to do it.

What happens if you make Selenium click on the option directly? Does it fail?

vikasmahato: I tried selecting the options directly and I observed  the following
1. It works fine on firefox with a glitch i.e. it is not selecting all options, rather it is revisitng some options which had previously been selected
2. It generates screenshots for the different options without actually selecting them hence leaving the select field blank (i.e. default selected option)

I have added all the relevant screenshots and the collect.py code here
https://gist.github.com/vikasmahato/58157977a4fc6d965ae6449fdbab37dc

collect.py
```diff
@@ -170,6 +172,10 @@ def do_something(driver, elem_attributes=None):
             driver.execute_script("arguments[0].value = '#ff0000'", elem)
         else:
             raise Exception('Unsupported input type: %s' % input_type)
+    elif elem.tag_name == 'select':
```
vikasmahato: The elem.click() was unnecessary here. I have removed it.

collect.py
marco-c: So the screenshots here are with the select click removed? It seems like we can't remove it then?

vikasmahato: No the screenshots were with using option directly. It appears that the elem.click() was unecessary and I have removed it and updated the pull request.

marco-c: Don't the screenshot show that in Chrome it doesn't work?

marco-c:
I'm a little confused @vikasmahato.
I've received via email this comment, but I see you have removed it:
> No the screenshots were with using option directly. It appears that the elem.click() was unecessary and I have removed it and updated the pull request.

Can you clarify the situation? Can you post screenshots with the current code?

collect.py
```diff
@@ -170,6 +172,10 @@ def do_something(driver, elem_attributes=None):
...
+        for option in elem.find_elements_by_tag_name('option'):
+            if option.text != '':
+                option.click()
```
marco-c: We don't want to click on all of them.

collect.py
```diff
@@ -170,6 +172,10 @@ def do_something(driver, elem_attributes=None):
             driver.execute_script("arguments[0].value = '#ff0000'", elem)
         else:
             raise Exception('Unsupported input type: %s' % input_type)
+    elif elem.tag_name == 'select':
+        for option in elem.find_elements_by_tag_name('option'):
```
marco-c: There's a bit of a problem. What if the order of the elements in the browsers is not the same?
We should click on exactly the same elements in both browsers, so we should make note of the attributes of the option element on which we have clicked on the first browser so that we can repeat the same action on the second browser.
Anyway, we can fix this in a follow-up.

vikasmahato:
> No the screenshots were with using option directly. It appears that the elem.click() was unecessary and I have removed it and updated the pull request.

This was regarding the screenshots I posted here https://gist.github.com/vikasmahato/58157977a4fc6d965ae6449fdbab37dc

In this gist I had tried to use the 'option' directly as you had once suggested. It worked in firefox but failed in chrome and hence I reverted to using 'select' followed by 'option'

I have also removed the redundant elem.click()  after
`elif elem.tag_name == 'select':` on line 175

The screenshots for the same are here:
Firefox
999990_0_firefox

Chrome
999990_0_chrome

collect.py
vikasmahato: I have added a break statement

vikasmahato:
@marco-c Just one more question! Do we keep track of autowebcompat issues on bugzilla?

marco-c:
> @marco-c Just one more question! Do we keep track of autowebcompat issues on bugzilla?

No, as issues on GitHub.

PR #102, opened 2018-03-04
Added test case for 'balance' function in utils.py for issue #69

### LINKED ISSUE #69: Write unit test for the 'balance' function in utils.py
A unit test has to be written for the `balance` function in utils.py.

See also PR #63.

adbugger:
Fixes #69.

As per my current understanding, the 'balance' function accepts a sequence of items (each item has label at index 1) and yields data items with alternating labels. The first label to be returned is the label that occurs first in the original sequence

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -71,7 +71,6 @@ def test_balance():
     assert(('data2', 1) == next(balanced_data))
     assert(('data4', 0) == next(balanced_data))
     assert(('data6', 1) == next(balanced_data))
-    assert(('data5', 0) == next(balanced_data))
 
     with pytest.raises(StopIteration):
         next(balanced_data)
@@ -89,9 +88,9 @@ def test_balance_unbalanced_data():
 
     assert(('data1', 1) == next(balanced_data))
     assert(('data4', 0) == next(balanced_data))
-    assert(('data2', 1) == next(balanced_data))
-    assert(('data5', 0) == next(balanced_data))
     assert(('data3', 1) == next(balanced_data))
+    assert(('data5', 0) == next(balanced_data))
+    assert(('data2', 1) == next(balanced_data))
 
     with pytest.raises(StopIteration):
         next(balanced_data)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -75,3 +75,23 @@ def test_balance():
 
     with pytest.raises(StopIteration):
         next(balanced_data)
+
+
+def test_balance_unbalanced_data():
+    unbalanced_tuples = [
+        ("data1", 1),
+        ("data2", 1),
+        ("data3", 1),
+        ("data4", 0),
+        ("data5", 0)]
+
+    balanced_data = utils.balance(unbalanced_tuples)
+
+    assert(('data1', 1) == next(balanced_data))
+    assert(('data4', 0) == next(balanced_data))
+    assert(('data3', 1) == next(balanced_data))
+    assert(('data5', 0) == next(balanced_data))
+    assert(('data2', 1) == next(balanced_data))
+
+    with pytest.raises(StopIteration):
+        next(balanced_data)
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -89,9 +89,9 @@ def test_balance_unbalanced_data():
 
     assert(('data1', 1) == next(balanced_data))
     assert(('data4', 0) == next(balanced_data))
-    assert(('data3', 1) == next(balanced_data))
-    assert(('data5', 0) == next(balanced_data))
     assert(('data2', 1) == next(balanced_data))
+    assert(('data5', 0) == next(balanced_data))
+    assert(('data3', 1) == next(balanced_data))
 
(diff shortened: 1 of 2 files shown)


tests/test_utils.py
```diff
@@ -52,3 +52,25 @@ def test_write_labels():
...
+    balanced_data = utils.balance(unbalanced_tuples)
+
+    try:
+        expected_label = 1
+        while True:
+            element = next(balanced_data)
+            assert(element[1] == expected_label)
+            expected_label = (expected_label + 1) % 2
+    except StopIteration:
+        del(balanced_data)
```
marco-c: Why the `del`? I think it can be removed.

tests/test_utils.py
```diff
@@ -52,3 +52,25 @@ def test_write_labels():
...
+        pass
```
marco-c: Assert the number of iterations.

tests/test_utils.py
adbugger: I thought it was a good idea to explicitly clean up the test environment. Will remove.

tests/test_utils.py
adbugger: Will do.

tests/test_utils.py
marco-c: In Python there's no need to do that, as the garbage collector takes care of it.

tests/test_utils.py
adbugger: How many iterations are there supposed to be? I have 3 '0' labels and 3 '1' labels. That should add up to a total of 6 iterations right?
The balance function is only yielding 5 values.

adbugger:
How many iterations are there supposed to be? I have 3 '0' labels and 3 '1' labels. That should add up to a total of 6 iterations right?
The balance function is only yielding 5 values.

marco-c:
> How many iterations are there supposed to be? I have 3 '0' labels and 3 '1' labels. That should add up to a total of 6 iterations right?
The balance function is only yielding 5 values.

Yes, it's yielding 5 values because it's not emptying the queue at the end. We should fix it, but in a separate PR.

Now that I think more about it, it would be better to be more explicit and check exactly what is yielded like so:
```Python
unbalanced_tuples = [
    ("data1", 1),
    ("data2", 1),
    ("data3", 0),
    ("data4", 0),
    ("data5", 0),
    ("data6", 1)]

balanced_data = balance(unbalanced_tuples)

assert(('data1', 1) == next(balanced_data))
assert(('data3', 0) == next(balanced_data))
assert(('data2', 1) == next(balanced_data))
assert(('data4', 0) == next(balanced_data))
assert(('data6', 1) == next(balanced_data))
assert(('data5', 0) == next(balanced_data))

with pytest.raises(StopIteration):
    next(balanced_data)
```

Also add a test where the data cannot be balanced, e.g.:
```Python
unbalanced_tuples = [
    ("data1", 1),
    ("data2", 1),
    ("data3", 1),
    ("data4", 0),
    ("data5", 0)]
```

adbugger:
Added the explicit checking of returned items. As expected, the item 'data5' is not yielded and we get a premature StopIteration exception.

I have some experience contributing to the Mozilla python codebase and would like to try fixing the actual balance function. The logic for what the function should do is fairly simple, except in the case where the data cannot be balanced. May I open another bug for that?

adbugger:
Added test case where data is inherently unbalanced. The order of data items with same labels is not preserved, but that is to be expected since we are technically using a stack, and not a queue. The test case has been written keeping this in mind, and it passes. Do tell if the relative order of items also needs to be maintained.

marco-c:
> I have some experience contributing to the Mozilla python codebase and would like to try fixing the actual balance function. The logic for what the function should do is fairly simple, except in the case where the data cannot be balanced. May I open another bug for that?

Yes, thanks! Create a new issue to do it (and when you submit a PR, fix both the function and the test).

> The order of data items with same labels is not preserved, but that is to be expected since we are technically using a stack, and not a queue. The test case has been written keeping this in mind, and it passes. Do tell if the relative order of items also needs to be maintained.

It shouldn't be a requirement, but to make it clearer let's use a queue instead of a stack. As for the other, file an issue and submit a separate PR to fix it.

marco-c:
I'd like to merge this PR first though, so please fix the test and I'll merge it. Then you can fix the function in a separate PR.

adbugger:
2 tests are now failing. I'll open another bug and fix both the function and the test in that PR.

tests/test_utils.py
```diff
@@ -52,3 +53,44 @@ def test_write_labels():
...
+    balanced_data = utils.balance(unbalanced_tuples)
+
+    assert(('data1', 1) == next(balanced_data))
+    assert(('data3', 0) == next(balanced_data))
+    assert(('data2', 1) == next(balanced_data))
+    assert(('data4', 0) == next(balanced_data))
+    assert(('data6', 1) == next(balanced_data))
+
+    with pytest.raises(StopIteration):
+        assert(('data5', 0) == next(balanced_data))
```
marco-c: This should only be `next(balanced_data)`.

PR #105, opened 2018-03-05
Added Resnet to the Networks 

Shashi456:
@marco-c 
The Resnet50 won the Imagenet competition once , i've added the same to the networks.py file. 

https://arxiv.org/abs/1512.03385

Taking hints from the original keras code [here](https://github.com/keras-team/keras/pull/3266/files).


Shashi456:
@marco-c okay.

Shashi456:
@marco-c  can this be reviewed now ?

marco-c:
It might be worth using the code from https://github.com/keras-team/keras/blob/master/examples/cifar10_resnet.py#L161, it could be more readable and allow us to use resnet v2 more easily in the future.

Shashi456:
@marco-c  i'm closing this pull request and will open a new one with the updated resnet.

PR #107, opened 2018-03-07
data-inconsistencies.py fixed

### LINKED ISSUE #78: Update data_inconsistencies.py script to account for changes in screenshot naming structure
#41 broke the data_inconsistencies.py script. We should modify it to make it work again.

Shashi456:
Fixes #78.


data_inconsistencies.py
```diff
@@ -39,7 +52,7 @@ def main():
     incons = get_inconsistencies()
     print('[*] {} inconsistencies found.'.format(len(incons)))
     print('[*] Writing to inconsistencies.csv... ', end='')
-    with open('inconsistencies.csv', 'w') as csvfile:
+    with open('inconsistencies.csv', 'w',encoding="utf-8") as csvfile:
```
marco-c: Why is this needed?

data_inconsistencies.py
```diff
@@ -17,8 +18,8 @@ def get_inconsistencies():
             sequence = int(parts[-2])
             element = '_'.join(parts[1:-2])
         else:
-            element = None
```
marco-c: Why did you change the default values of the variables?

PR #108, opened 2018-03-07
Rewrote the 'balance' function in utils.py to fix issue #106

### LINKED ISSUE #106: utils.py 'balance' function does not return last element in balanced data
According to the discussions with @marco-c on issue #69 , the balance function needs to be re-written:
1. to ensure that we return the last element in case of balanced data. This will be done by emptying the queue upon exit
2. to use queue implementation to balance, instead of stacks. This is done to preserve relative ordering amongst data items with same labels

Once that PR is closed, I'd be happy to start work on this bug. This bug will update the 'balance' function as well as its test cases.

adbugger:
Fixes #106.

The balance function has now been rewritten to maintain relative order of elements, and return all the elements in a balanced data list. The test cases have also been rewritten to reflect this fact.


autowebcompat/utils.py
```diff
@@ -140,32 +140,40 @@ def next(self):
 
 
 def balance(it):
-    last_label = None
-    queue_1 = []
-    queue_0 = []
+    # Initialise last_label to not (label of first element)
+    last_label = it[0][1]
```
marco-c: The function needs to support an iterator, `it` could be a list or an iterator.
You should add a test case that passes to balance an iterator, so you can make sure it's working in both cases.

adbugger:
I realised that the initialisation of `last_label` to inverse of the first label was not needed. Changed both the `balance` function and added a test case for a list iterator.

tests/test_utils.py
```diff
@@ -88,9 +89,32 @@ def test_balance_unbalanced_data():
...
     assert(('data1', 1) == next(balanced_data))
     assert(('data4', 0) == next(balanced_data))
-    assert(('data3', 1) == next(balanced_data))
+    assert(('data2', 1) == next(balanced_data))
     assert(('data5', 0) == next(balanced_data))
+    assert(('data3', 1) == next(balanced_data))
+
+    with pytest.raises(StopIteration):
+        next(balanced_data)
+
```
marco-c: Add the iterator test in the same function, so you don't have to define the data twice. You could also reuse the assertions if you do something like this:
```Python
for input_type in [list, iter]:
    balanced_data = utils.balance(input_type(unbalanced_tuples))
    ...
```

marco-c: There should be even a pytest way to parametrize the function, if you want.

autowebcompat/utils.py
```diff
@@ -140,32 +140,40 @@ def next(self):
...
-                    e = queue_0.pop()
-                else:
-                    if len(queue_1) == 0:
-                        break
-                    e = queue_1.pop()
-
-                last_label = e[1]
-                yield e
+            queue[cur_label].append(e)
+            other_label = 1 if cur_label == 0 else 0
```
marco-c: I think this like and the following three lines are unneeded

tests/test_utils.py
```diff
@@ -55,22 +55,26 @@ def test_write_labels():
...
 
-def test_balance():
-    unbalanced_tuples = [
-        ("data1", 1),
-        ("data2", 1),
-        ("data3", 0),
-        ("data4", 0),
-        ("data5", 0),
-        ("data6", 1)]
+data = [
```
marco-c: Nit: Call this `unbalanced_data` or `test_balance_data` as `data` is too generic.

PR #111, opened 2018-03-09
Make sure there's only one element on the page with the given attributes

### LINKED ISSUE #86: Make sure there's only one element on the page with the given attributes
In the first browser, when the crawler finds an element, it should check if there's no other element on the page with the same attributes.

In the second browser, when we are repeating the steps, we should assert that there's a single element on the page with the given attributes.

ManasviKundalia:
Fixes #86.


ManasviKundalia:
@marco-c please [review](https://github.com/marco-c/autowebcompat/pull/111/commits/779fb2aeffe6fb200bda1a6df9a7ac5c3281b12a)

collect.py
```diff
@@ -119,25 +150,24 @@ def do_something(driver, elem_attributes=None):
...
+            if len(res) == 1:
+                elem = child
+                break
+            else:
+                # remove all elements with the same attributes
+                print('Found elements with same attributes')
+                # print(res)
+                print()
+                for element in res:
+                    children.remove(element)
```
sagarvijaygupta: You should never remove elements from the list you are iterating on. You can iterate over a copy and remove elements from the original.
```
lst = [1, 2, 1, 1, 3, 1, 1]
for elem in lst:
    if elem == 1:
        lst.remove(elem)
print(lst)
```
This will result in output as [2,3,1,1] and not [2,3]

collect.py
```diff
@@ -119,25 +150,24 @@ def do_something(driver, elem_attributes=None):
             if not child.is_displayed() or not child.is_enabled():
                 continue
 
-            elem = child
-            break
+            res = check_attributes(driver, elem_attributes, child)  # returns list of elements with attributes = child_attributes
```
marco-c: Pass `children` as a parameter to this function, so you don't have to find the elements again.

collect.py
marco-c: Don't remove anything from the list, but define a list of elements to ignore and put them there.

collect.py
```diff
@@ -95,6 +95,37 @@ def get_all_attributes(driver, child):
     return child_attributes
 
 
+def check_attributes(driver, elem_attributes, elem=None):
```
marco-c: Rename this function to something like `get_elements_with_attributes` and make it return a list of elements matching the attributes.
The callers will check the length of the list and make sure it is 1.

ManasviKundalia:
@marco-c I did the necessary changes. Please review.
Also the website http://www.danielhopwood.com/ has a number of elements with same attributes. You could use it for testing.

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
     return child_attributes
 
 
+def get_elements_with_attributes(driver, elem_attributes, children):
+
```

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
...
+    same_attributes = []
```
marco-c: The name of this variable is a bit confusing, can you find a better name?

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
...
+    for child in children:
+        child_attributes = get_all_attributes(driver, child)
+        if child_attributes == elem_attributes:
+            same_attributes.append(child)
+    return same_attributes
+
+
 def do_something(driver, elem_attributes=None):
     elem = None
+    body = driver.find_elements_by_tag_name('body')
```
marco-c: Nit: newline before this line

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
...
+    assert len(body) == 1
+    body = body[0]
+
+    buttons = body.find_elements_by_tag_name('button')
+    links = body.find_elements_by_tag_name('a')
+    inputs = body.find_elements_by_tag_name('input')
+    selects = body.find_elements_by_tag_name('select')
+    children = buttons + links + inputs + selects
```
marco-c: Nit: newline after this line

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
...
     if elem_attributes is None:
-        body = driver.find_elements_by_tag_name('body')
-        assert len(body) == 1
-        body = body[0]
-
-        buttons = body.find_elements_by_tag_name('button')
-        links = body.find_elements_by_tag_name('a')
-        inputs = body.find_elements_by_tag_name('input')
-        selects = body.find_elements_by_tag_name('select')
-        children = buttons + links + inputs + selects
```

collect.py
```diff
@@ -95,22 +95,36 @@ def get_all_attributes(driver, child):
...
 
         random.shuffle(children)
+        list_ignore = []  # list of elements with same attributes to ignore
```
marco-c: Use a more descriptive name for this variable. E.g. `children_to_ignore`

collect.py
```diff
@@ -119,25 +133,21 @@ def do_something(driver, elem_attributes=None):
             if not child.is_displayed() or not child.is_enabled():
                 continue
 
-            elem = child
-            break
+            res = get_elements_with_attributes(driver, elem_attributes, children)  # returns list of elements with attributes = child_attributes
```
marco-c: `res` is a confusing name, please use something better (e.g. `elems`)

marco-c: Also remove the comment, as the function call is self-explanatory.

collect.py
```diff
@@ -119,25 +133,21 @@ def do_something(driver, elem_attributes=None):
...
+            if len(res) == 1:
+                elem = child
+                break
+            else:
+                # remove all elements with the same attributes
+                print('Found elements with same attributes')
+                # print(res)
```
marco-c: Remove these logging statements.

collect.py
```diff
@@ -119,25 +133,21 @@ def do_something(driver, elem_attributes=None):
...
-            inputs = body.find_elements_by_tag_name('input')
-            selects = body.find_elements_by_tag_name('select')
-            children = buttons + links + inputs + selects
-
-            for child in children:
-                # Get all the attributes of the child.
-                if elem_attributes == get_all_attributes(driver, child):
-                    elem = child
-                    break
+            res = get_elements_with_attributes(driver, elem_attributes, children)  # returns elem if unique
```
marco-c: Remove the comment as it is self-explanatory and rename `res` with a better name.

ManasviKundalia:
@marco-c Sorry for the delay! I did the changes. Please review.

collect.py
```diff
@@ -15,7 +15,7 @@
 MAX_THREADS = 5
 
 if sys.platform.startswith("linux"):
-    chrome_bin = "tools/chrome-linux/chrome"
+    chrome_bin = "/home/manasvi/Documents/tools/chrome-linux/chrome"
```
marco-c: Please revert this

collect.py
```diff
@@ -95,22 +95,37 @@ def get_all_attributes(driver, child):
...
+    assert len(body) == 1
+    body = body[0]
+
+    buttons = body.find_elements_by_tag_name('button')
+    links = body.find_elements_by_tag_name('a')
+    inputs = body.find_elements_by_tag_name('input')
+    selects = body.find_elements_by_tag_name('select')
+    children = buttons + links + inputs + selects
+
+    if elem_attributes is None:
```
marco-c: Nit: sorry I missed this earlier, this newline should be removed here (there should be no newlines after `if`, `for`, etc.).

ManasviKundalia:
@marco-c any other changes? 
I mean, the codecov checks weren't successful.. what should I do about that?

marco-c:
>  I mean, the codecov checks weren't successful.. what should I do about that?

No worries about that, there are no tests for the crawler yet, so it's fine!
Thanks!

ManasviKundalia:
@marco-c Welcome! :)

PR #120, opened 2018-03-12
Add flake8-isort

### LINKED ISSUE #119: Use flake8-isort

vrishank97:
Fixes #119.


autowebcompat/utils.py
```diff
@@ -1,12 +1,14 @@
+import csv
 import json
 import os
 import random
 import threading
+
+import keras
 import numpy as np
+from keras.preprocessing.image import (ImageDataGenerator, img_to_array,
```
marco-c: We should set it up so that keras appears before numpy

collect.py
```diff
@@ -1,16 +1,19 @@
...
+import glob
 import json
 import os
+import random
 import sys
 import time
-import random
 import traceback
-import glob
+from concurrent.futures import ThreadPoolExecutor
```
marco-c: Same here, I'd like concurrent.futures to be before glob

collect.py
```diff
@@ -1,16 +1,19 @@
...
+
 from PIL import Image
-from selenium import webdriver
-from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
 from autowebcompat import utils
+from selenium import webdriver
+from selenium.common.exceptions import (NoAlertPresentException,
```
marco-c: Let's setup a longer maximum length, as I prefer one line to multiple lines

generate_labels.py
```diff
@@ -2,7 +2,6 @@
 
 from autowebcompat import utils
 
-
```
marco-c: Is this needed?

requirements-dev.txt
```diff
@@ -0,0 +1,6 @@
+Pillow==4.0.0
```
marco-c: There's already a requirements-test.txt file that is serving the same purpose, can you update that one and remove this?

vrishank97:
@marco-c I've made the changes requested

.isort.cfg
```diff
@@ -0,0 +1,7 @@
+[settings]
+known_numpy=numpy
```
marco-c: Is this the only way to acheve that?
What if we set the force_alphabetical_order instead?

.isort.cfg
```diff
@@ -0,0 +1,7 @@
...
+known_keras=keras
+known_concurrent=concurrent.futures
+sections=CONCURRENT,FUTURE,STDLIB,KERAS,NUMPY,THIRDPARTY,FIRSTPARTY,LOCALFOLDER
```
marco-c: What is the `sections` parameter doing? Can you point me to the documentation where you found it?

vrishank97:
It sets the order of imports

Link : https://pypi.python.org/pypi/isort under Custom Sections and Ordering

.isort.cfg
```diff
@@ -0,0 +1,7 @@
...
+no_lines_before=NUMPY
```
marco-c: Why no lines before numpy?

requirements.txt
```diff
@@ -2,4 +2,4 @@ Pillow==4.0.0
 selenium==3.4.3
 keras==2.1.1
 tensorflow==1.3.0
-h5py==2.7.1
+h5py==2.7.1
```
marco-c: Please avoid extraneous changes

.isort.cfg
vrishank97: By default the order is for section that are separated with black lines, here I assumed we wanted them in the same section without separation.

.isort.cfg
marco-c: > It sets the order of imports
> Link : https://pypi.python.org/pypi/isort under Custom Sections and Ordering

OK, we definitely don't want this. Why should concurrent always come before anything else?

vrishank97: I'll look into ways of creating import exceptions for specific files.

requirements.txt
vrishank97: Oops

.isort.cfg
marco-c: Why creating exceptions for specific files?

vrishank97: The same config is going to be used for the entire repo. So if we don't want to set concurrent.future as the highest priority import globally, we'll need to set it up as an exception.

marco-c: To be clear, what I'm asking is not about concurrent.futures specifically.
```
from concurrent.futures import ThreadPoolExecutor
import glob
```
and not:
```
import glob
from concurrent.futures import ThreadPoolExecutor
```

OR

```
from aSTANDARDLIBRARY import SOMETHING
import glob
```
and not:
```
import glob
from aSTANDARDLIBRARY import SOMETHING
```

We want to do this consistently without using exceptions, it should be the default for the entire project and not only for concurrent.futures but for anything else too.

vrishank97: I can set the priority for standard libraries to be higher than that for glob

marco-c: `glob` is a standard library too. I don't want to make glob always appear after other standard libraries. I want `glob` to appear after `concurrent.futures` because `glob` starts with `g` and `concurrent.futures` starts with `c`. But `glob` should still appear before other standard libraries whose name starts with a letter > `g`.

marco-c:
@vrishank97 to clarify again, we don't want to add exceptions, as they are not maintainable.
We want the packages to be in alphabetical order. The priority for `glob` should not be changed.

vrishank97:
Oops. I had pushed the changes before reading this. So we just need the imports for each section to be ordered alphabetically. Right?

marco-c:
> Oops. I had pushed the changes before reading this. So we just need the imports for each section to be ordered alphabetically. Right?

Yes :)
So `import glob` should come after `from concurrent.futures import ThreadPoolExecutor` if you manage to find a way to do it (**without using exceptions** as we want this behavior for all packages).

vrishank97:
Figured out the problem. Currently isort arranges imports for each section alphabetically, however it separates lines with different forms of import statement syntax.
Lines with `"from" module "import" identifier` are separated from lines with `"import" module`.
I'll look into ways of working around this.

.isort.cfg
vrishank97: It would work but it won't separate numpy imports from keras imports

marco-c: We don't want to separate them, they are both third-party libraries.

.isort.cfg
vrishank97: Oops. I assumed you wanted keras and numpy to be separated.

marco-c: No, I like the default choice: standard libraries first, then third-party libraries, then our libraries.

vrishank97: The latest commit uses the default ordering, however the issue with concurrent is still there.

marco-c:
>  I'll look into ways of working around this.

Did you manage to find a way to do this?

vrishank97:
Yeah. I used force_sort_within_sections=True to force alphabetical sorting within sections. Are there any other changes required?

marco-c:
Just two nits: generate_labels.py and requirements.txt contain unrelated changes, can you undo them?
Also, there are conflicts to resolve.

marco-c:
Ah, the conflicts need to be resolved before merging.

autowebcompat/utils.py
```diff
@@ -1,12 +1,13 @@
...
 import os
 import random
 import threading
-import numpy as np
+
 from PIL import Image
 import keras
-from keras.preprocessing.image import ImageDataGenerator, load_img, img_to_array
-import csv
+from keras.preprocessing.image import ImageDataGenerator, img_to_array, load_img  # NOQA
```
marco-c: Ah, I didn't notice this before, why the exception here?

collect.py
```diff
@@ -1,16 +1,17 @@
...
 import traceback
-import glob
+
 from PIL import Image
-from selenium import webdriver
-from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
 from autowebcompat import utils
+from selenium import webdriver
+from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException  # NOQA
```
marco-c: Why the exception here?

data_inconsistencies.py
```diff
@@ -1,8 +1,10 @@
-import os
 import csv
+import os
+
 from autowebcompat import utils
 
 
+
```
marco-c: This seems unneeded?

.isort.cfg
```diff
@@ -0,0 +1,3 @@
+[settings]
+force_sort_within_sections=True
+force_single_line=True
```
marco-c: We don't want to force one import per line.

vrishank97:
I have removed force_single_line and set line_length=500

marco-c:
500 is a bit too much, please use a more reasonable value (say, 150).
It looks like Travis is failing.

vrishank97:
'tempfile.TemporaryDirectory' is being imported but is unused in test_utils.py and this is making the Travis build fail. Should I remove it?

marco-c:
> 'tempfile.TemporaryDirectory' is being imported but is unused in test_utils.py and this is making the Travis build fail. Should I remove it?

The import has been removed already, probably you brought it back when resolving conflicts.

vrishank97:
Travis shows
./collect.py:11:1: I004 isort found an unexpected blank line in imports
./collect.py:12:1: I001 isort found an import in the wrong position
./collect.py:15:1: I003 isort expected 1 blank line in imports, found 0

However there are no errors when I run isort on collect.py on my machine. Can you help me out?

test-requirements.txt
```diff
@@ -1,3 +1,4 @@
 flake8
```
marco-c: What version of flake8 are you using locally? Maybe try updating your version of flake8 and pin the version here, so you are sure you have the same version as on Travis.

Also make sure you are running everything with Python 3

generate_labels.py
marco-c: This seems unneeded

test-requirements.txt
vrishank97: I'm on flake8 3.5.0 and everything runs on Python3

marco-c:
I see the same errors as Travis locally.
Here's the output of `flake8 --version` for me: `3.5.0 (flake8_isort: 2.3, mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 3.6.3 on Linux`.

vrishank97:
My output for `flake8 --version` is `3.5.0 (flake8_isort: 2.3, mccabe: 0.6.1, pycodestyle: 2.3.1, pyflakes: 1.6.0) CPython 3.6.3 on Darwin`
I'll try reinstalling flake8.

marco-c:
Really weird, you have exactly the same version.
If you can't get it to work, try to fix the issues that Travis is telling you to fix by using the Travis log.

vrishank97:
I'm getting the same output as travis on my google cloud compute vm. There might be some mac related bug.

marco-c:
Good, so you can fix the two remaining issues more easily.

vrishank97:
Here is the flake8 output from my mac
screen shot 2018-03-15 at 11 27 32 pm
This is the output from a linux machine
screen shot 2018-03-15 at 11 27 43 pm

vrishank97:
The order of imports is the same as the second commit which was successful on travis, except for the line length and concurrent.futures being on top.

collect.py
```diff
@@ -1,16 +1,17 @@
...
-import random
 import traceback
-import glob
+
 from PIL import Image
-from selenium import webdriver
-from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
 from autowebcompat import utils
+from selenium import webdriver
```
marco-c: The selenium imports are third-party libraries, so they should go in the second section of the imports. This is probably the cause of at least one of the flake8 failures.

vrishank97:
Finally 😄. This took a lot longer than I expected...

PR #122, opened 2018-03-13
Added tag support for crawler

### LINKED ISSUE #104: Use tag name too to identify elements
Right now we use the attributes, it would be more precise if we add the tag name too.

Trion129:
Fixes #104. 
Adds tag name to the list of attributes for the element.

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -93,11 +93,12 @@ def get_element_properties(driver, child):
         elem_properties.attributes[arguments[0].attributes[i].name] = arguments[0].attributes[i].value;
       }
       elem_properties.tag = arguments[0].tagName;
-      
+
       return elem_properties;
     """, child)
 
     return child_properties
+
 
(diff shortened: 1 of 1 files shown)


collect.py
```diff
@@ -89,6 +89,8 @@ def get_all_attributes(driver, child):
       for (let i = 0; i < arguments[0].attributes.length; i++) {
         elem_attribute[arguments[0].attributes[i].name] = arguments[0].attributes[i].value;
       }
+      elem_attribute["tag"] = arguments[0].tagName;
```
marco-c: What if one of the attributes is called "tag"?

Trion129: oops 😋 yeah, that's a possibility, will fix it

Trion129:
Instead of attributes, now I use properties term, it is a dict of format
```
{
  "tag",
  "attributes"
}
```

😂 My branch keeps falling out of sync

PR #125, opened 2018-03-14
comparing images with overlayed at alpha =0.5 for labelling

amit2rockon:
Y for  y
D for d
N for n
Through this way we can compare images at same time and can see difference in area where they differ .


amit2rockon:
@marco-c  can we think over labeling in this way?

PR #127, opened 2018-03-14
Data Inconsistency metrics 

### LINKED ISSUE #87: In data_inconsistencies.py, print how many screenshots we have collected and how many we could have collected
@marco-c 
Once #78  is fixed, I think a script should be written which ensures that inconsistent data is excluded and/or what part of the data collected is actually being used to train so that we have an idea of how successful we are at data collection and maybe update the crawler script later on.
Although i do know that labels.py will only take a set of images if both are present, i haven't seen any enforcement measures.

Shashi456:
These are some inconsistency metrics on the dataset.
Fixes #87.


Shashi456:
@marco-c 

``` bash

Number of Photos: 16742
Number of Pair of images: 5676
Number of Pair of images possible: 11066
Percentage of Firefox Inconsistencies:0
Percentage of Chrome Inconsistencies:100
```
The output looks something like that. One interesting thing to observe is we have all inconsistencies present because of chrome, can you think of any reasons?

data_inconsistencies.py
```diff
@@ -32,6 +33,35 @@ def get_inconsistencies():
     return incons
 
 
+def inconsistencies_metric(file_name):
+    f = utils.get_all_images()
+    total_img = len(f)
+    with open(file_name, 'rU') as csvfile:
```
marco-c: You are re-reading the file that is created with this script, it would be better to use the data that we have generated with `get_inconsistencies` directly

data_inconsistencies.py
```diff
@@ -32,6 +33,35 @@ def get_inconsistencies():
     return incons
 
 
+def inconsistencies_metric(file_name):
```
marco-c: Rename the function to `print_statistics` (inconstincies is implied by the file name)

data_inconsistencies.py
```diff
@@ -32,6 +33,35 @@ def get_inconsistencies():
...
+                except KeyError:
+                    data[header] = [value]
+
+    firefox = data['FIREFOX']
+    n_incons = len(firefox)
+    chrome = data['CHROME']
+    incons_f = [int(x == "True") for x in firefox]
+    incons_c = [int(x == "True") for x in chrome]
+    print("Number of Photos: %d " % total_img)
+    print("Number of Pair of images: %d " % int((total_img-n_incons)/2))
```
marco-c: Shouldn't it be `(total_img - n_incons * 2) / 2`?

data_inconsistencies.py
```diff
@@ -32,6 +33,35 @@ def get_inconsistencies():
...
+                    data[header].append(value)
+                except KeyError:
+                    data[header] = [value]
+
+    firefox = data['FIREFOX']
+    n_incons = len(firefox)
+    chrome = data['CHROME']
+    incons_f = [int(x == "True") for x in firefox]
+    incons_c = [int(x == "True") for x in chrome]
+    print("Number of Photos: %d " % total_img)
```
marco-c: Nit: use .format instead of %

data_inconsistencies.py
marco-c: Nit: here and in the following, put spaces between the math operators to make the formulas easier to read

data_inconsistencies.py
```diff
@@ -1,5 +1,6 @@
 import os
 import csv
+from autowebcompat import utils
 
 
 def get_inconsistencies():
```
marco-c: This function should use utils.get_all_images() too, but it should be changed in a separate PR.
Can you do it?

marco-c:
> The output looks something like that. One interesting thing to observe is we have all inconsistencies present because of chrome, can you think of any reasons?

The reason is that the crawler starts with Firefox. So if there's something that we can't do in Firefox, it won't be done in Chrome either.

data_inconsistencies.py
Shashi456: Total images - n_icons because in the photos which are inconsistent , there is no paired photo . So n_incons *2 would be wrong.

data_inconsistencies.py
Shashi456: @marco-c  wouldn't it just be writing ``` files = utils.get_all_images() ``` instead of ```listdir(data)``` ? or is there a corner case to think about ?

Shashi456:
@marco-c  one thing to keep in mind is now that i'm using get_all_images() i'm getting more inconsistencies then listdir(), and having checked a few files compared to earlier version none of them are false inconsistencies.

data_inconsistencies.py
marco-c: Yes, just that

marco-c: It should be done in a separate PR.

marco-c:
> @marco-c one thing to keep in mind is now that i'm using get_all_images() i'm getting more inconsistencies then listdir(), and having checked a few files compared to earlier version none of them are false inconsistencies.

This is a bit unexpected. Can you list a few examples of inconsistencies found with get_all_images() that are not found with listdir()? (Please reply to this question in the separate PR where you will do this change)

data_inconsistencies.py
```diff
@@ -32,17 +32,35 @@ def get_inconsistencies():
...
+    firefox = []
+    chrome = []
+    print('[*] {} inconsistencies found.'.format(n_incons))
     print('[*] Writing to inconsistencies.csv... ', end='')
     with open('inconsistencies.csv', 'w') as csvfile:
         writer = csv.writer(csvfile, delimiter=',')
         writer.writerow(['WEBCOMPAT-ID', 'SEQUENCE-NO', 'FIREFOX', 'CHROME'])
         for line in incons:
             writer.writerow(line)
+            firefox.append(line[2])
```
marco-c: Do this in the print_statistics function, as it's unrelated to writing the file.

marco-c: That is, pass `incons` to print_statistics

data_inconsistencies.py
```diff
@@ -32,17 +32,35 @@ def get_inconsistencies():
     return incons
 
 
+def print_statistics(file_name, n_incons, firefox, chrome):
+    f = utils.get_all_images()
+    total_img = len(f)
+    incons_f = [int(x) for x in firefox]
+    incons_c = [int(x) for x in chrome]
+    print("Number of Photos: {} " .format(total_img))
```
marco-c: Nit: uppercase only for the first letter and names (in this case `Number of photos`)

data_inconsistencies.py
```diff
@@ -32,17 +32,35 @@ def get_inconsistencies():
...
+    print("Number of Pair of images: {} " .format(int((total_img - n_incons) / 2)))
```
marco-c: Nit: use the plural here `pairs` instead of `pair`

Shashi456:
@marco-c  i will file a seperate pr as soon as this is merged to avoid conflicts since its on the same file.

data_inconsistencies.py
```diff
@@ -32,17 +33,37 @@ def get_inconsistencies():
     return incons
 
 
+def print_statistics(file_name, incons, n_incons):
+    f = utils.get_all_images()
+    total_img = len(f)
+    firefox = []
```
marco-c: Define n_incons here since you have incons here so the signature of the function will be cleaner

data_inconsistencies.py
```diff
@@ -32,17 +33,37 @@ def get_inconsistencies():
...
+    for line in incons:
+        firefox.append(line[2])
+        chrome.append(line[3])
+
+    incons_f = [int(x) for x in firefox]
+    incons_c = [int(x) for x in chrome]
+    print("Number of photos: {} " .format(total_img))
+    print("Number of pairs of images: {} " .format(int((total_img - n_incons) / 2)))
+    print("Number of pairs of images possible: {} " .format(int((total_img - n_incons) / 2 + n_incons)))
+    print("Percentage of Firefox Inconsistencies: {}  " .format(int(((n_incons - sum(incons_f)) / n_incons) * 100)))
```
marco-c: Nit: inconsistencies should be lower case

sagarvijaygupta:
>@marco-c one thing to keep in mind is now that i'm using get_all_images() i'm getting more inconsistencies then listdir(), and having checked a few files compared to earlier version none of them are false inconsistencies.

@marco-c  @Shashi456 this is because when we use listdir() even text files are included in count which adds the `webcompat_id` as a `browser` name which works as a disguise for `chrome`

Shashi456:
@sagarvijaygupta i agree i knew some files were being hidden but hadnt pinpointed which ones , thanks.

data_inconsistencies.py
```diff
@@ -32,17 +33,38 @@ def get_inconsistencies():
...
+    print("Number of pairs of images possible: {} " .format(int((total_img - n_incons) / 2 + n_incons)))
+    print("Percentage of Firefox inconsistencies: {}  " .format(int(((n_incons - sum(incons_f)) / n_incons) * 100)))
+    print("Percentage of Chrome inconsistencies: {} " .format(int(((n_incons - sum(incons_c)) / n_incons) * 100)))
+
+
 def main():
     print('[*] Getting inconsistencies in screenshots...')
     incons = get_inconsistencies()
-    print('[*] {} inconsistencies found.'.format(len(incons)))
+    n_incons = len(incons)
```
marco-c: This change is unneeded now

Issue #118, opened 2018-03-12
When performing labeling, allow user to select the area of the image where the incompatibility lies


marco-c:
With the current labeling system, we don't know where the incompatibilities lie but just whether there is at least an incompatibility or not.
This means we can't do things like training on clips of the images (because we wouldn't know if the clip contains an incompatibility or not).

We should store the areas where the incompatibilities lie alongside the labels.

sagarvijaygupta:
@marco-c Can I take this one? I should add a tkinter based interface for the same?

marco-c:
Sure! The current label.py can be modified, or if it is a pain to do it with tkinter you can use anything else.

sagarvijaygupta:
@marco-c In that case I will try to use OpenCV. Is it fine?

marco-c:
Yes, whatever you prefer/find easier to use.

marco-c:
The only requirement is that the resulting UI must be simple enough to use for labeling.

sagarvijaygupta:
What should be the labeling for screenshots which differ in language? Like this one

![screenshot from 2018-06-09 15-09-58](https://user-images.githubusercontent.com/19645459/41190153-57d69db4-6bf7-11e8-9c37-a37bad89e1b3.png)

marco-c:
What is the URL for this case?

marco-c:
I guess we should classify them as being different, since the language detection worked in one browser but not in the other.

sagarvijaygupta:
Okay! So I should mark the bounding box around whole text? And in the English side only?

marco-c:
> Okay! So I should mark the bounding box around whole text? And in the English side only?

Yes, around the text (and in this case the "NEXT" button too). In this case it looks like the English side is wrong, since the screenshots were collected on my machine.

sagarvijaygupta:
Oh! So in all images where there is language difference I will mark bounding boxes on Italian side.

marco-c:
I guess we can close this now, as it was done.

PR #126, opened 2018-03-14
comparing images with overlayed at alpha =0.5 for labelling

amit2rockon:
Through this way we can compare images at same time and can see difference in area where they differ . 
D for
d
N for
n
Y for
y

PR #128, opened 2018-03-14
Add draw_boundary_box to select area where incompatibility lies.

sagarvijaygupta:
Solves #118

How it works?
- Keypress of ```'n', 'y', ' d'``` works same as they worked previously.
- Press ```'m' ``` to mark the images as ```not compatible``` (ie same as what pressing ```n``` does) and enter into selection region.
- Select different boundary boxes (you can select more than one) in both images.
- If you make a mistake press ```'r'``` to reset all boundary boxes.
- Press ```Escape``` to exit selection region without saving your ```boundary_boxes```
- Press ```Return``` to save your ```boundary_boxes``` into a file in directory ```label_persons``` named as ```filename_boundary_box.csv```
- Data in ```filename_boundary_box.csv``` is stored in format ```filename,list _of_boundary_boxes```

### DIFF
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -198,20 +198,15 @@ def write_labels(labels, file_name='labels.csv'):
             writer.writerow([key, values])
 
 
-def read_boundary_boxes(file_name):
+def read_bounding_boxes(file_name):
     try:
         with open(file_name, 'r') as f:
-            next(f)
-            reader = csv.reader(f)
-            boundary_boxes = {row[0]: ast.literal_eval(row[1]) for row in reader}
+            bounding_boxes = json.load(f)
     except FileNotFoundError:
-        boundary_boxes = {}
-    return boundary_boxes
+        bounding_boxes = {}
+    return bounding_boxes
 
 
-def write_boundary_boxes(boundary_boxes, file_name):
+def write_bounding_boxes(bounding_boxes, file_name):
     with open(file_name, 'w') as f:
-        writer = csv.writer(f, delimiter=',')
-        writer.writerow(["Image Name", "Boundary Boxes (top_left_x,top_left_y,bottom_right_x,bottom_right_y)"])
-        for key, values in boundary_boxes.items():
-            writer.writerow([key, values])
+        print(json.dumps(bounding_boxes), file=f)
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -1,3 +1,4 @@
+import ast
 import json
 import os
 import random
@@ -194,4 +195,23 @@ def write_labels(labels, file_name='labels.csv'):
         writer = csv.writer(f, delimiter=',')
         writer.writerow(["Image Name", "Label"])
         for key, values in labels.items():
+            writer.writerow([key, values])
+
+
+def read_boundary_boxes(file_name):
+    try:
+        with open(file_name, 'r') as f:
+            next(f)
+            reader = csv.reader(f)
+            boundary_boxes = {row[0]: ast.literal_eval(row[1]) for row in reader}
+    except FileNotFoundError:
+        boundary_boxes = {}
+    return boundary_boxes
+
+
+def write_boundary_boxes(boundary_boxes, file_name):
+    with open(file_name, 'w') as f:
+        writer = csv.writer(f, delimiter=',')
+        writer.writerow(["Image Name", "Boundary Boxes (top_left_x,top_left_y,bottom_right_x,bottom_right_y)"])
+        for key, values in boundary_boxes.items():
             writer.writerow([key, values])
--- a/generate_labels.py
+++ b/generate_labels.py
@@ -11,6 +11,8 @@ ydn_map = {'y': 0, 'd': 1, 'n': 2}
 ydn_reverse_map = {0: 'y', 1: 'd', 2: 'n'}
 
 for file_name in all_file_names:
+    if 'boundary_box' in file_name:
+        continue
     labels = utils.read_labels(labels_directory + file_name)
     for key, value in labels.items():
         if key not in labels_voted.keys():
--- a/label.py
+++ b/label.py
@@ -1,6 +1,8 @@
 import argparse
 from PIL import ImageTk, Image
 from tkinter import Tk, Label
+import cv2
+import numpy as np
 
 
 from autowebcompat import utils
@@ -11,16 +13,52 @@ parser = argparse.ArgumentParser()
 parser.add_argument("file_name", action="store")
 args = parser.parse_args()
 
-labels = utils.read_labels(labels_directory + args.file_name)
+labels = utils.read_labels(labels_directory + args.file_name + ".csv")
+boundary_boxes = utils.read_boundary_boxes(labels_directory + args.file_name + "_boundary_box.csv")
 
 images_to_show = [i for i in utils.get_images() if i not in labels]
 current_image = None
+drawing = False
 
 root = Tk()
 panel1 = Label(root)
 panel1.pack(side="left", padx=10)
 panel2 = Label(root)
 panel2.pack(side="left", padx=10)
+
+
+def draw_boundary_box(event, mouse_x, mouse_y, flags, param):
+    global start_x, start_y, drawing, end_x, end_y
+    [drawing_area, boxes] = param
+
+    if event == cv2.EVENT_LBUTTONDOWN:
+        drawing = True
+        start_x, start_y = mouse_x, mouse_y
+        end_x, end_y = mouse_x, mouse_y
+    elif event == cv2.EVENT_MOUSEMOVE:
+        if drawing is True:
+            box_start_x, box_start_y = start_x, start_y
+            box_end_x, box_end_y = end_x, end_y
+            if box_start_x > box_end_x:
+                box_start_x, box_end_x = box_end_x, box_start_x
+            if box_start_y > box_end_y:
+                box_start_y, box_end_y = box_end_y, box_start_y
+            box_start_x = max(0, box_start_x)
+            box_start_y = max(0, box_start_y)
+            box_end_x = min(drawing_area.shape[1], box_end_x)
+            box_end_y = min(drawing_area.shape[0], box_end_y)
+            drawing_area[box_start_y:box_end_y + 1, box_start_x:box_end_x + 1, 0:3] = 0
+            end_x = max(mouse_x, 0)
+            end_y = max(mouse_y, 0)
+            cv2.rectangle(drawing_area, (start_x, start_y), (end_x, end_y), (0, 255, 0), -1)
+    elif event == cv2.EVENT_LBUTTONUP:
+        drawing = False
+        start_x = max(0, start_x)
+        start_y = max(0, start_y)
+        end_x = min(drawing_area.shape[1], end_x)
+        end_y = min(drawing_area.shape[0], end_y)
+        cv2.rectangle(drawing_area, (start_x, start_y), (end_x, end_y), (0, 255, 0), -1)
+        boxes.append([min(start_x, end_x), min(start_y, end_y), max(start_x, end_x), max(start_y, end_y)])
 
 
 def get_new_image():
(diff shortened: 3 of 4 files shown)


sagarvijaygupta:
screenshot from 2018-03-14 15-43-56

marco-c:
What about this:
- press `y` to mark compatible;
- press `d` to mark compatible but with differences, let the user select the regions and then press Enter when they are done;
- press `n` to mark incompatible, let the user select the regions and then press Enter when they are done.

This way we don't have too many controls and we force the user to select the incompatible regions.

marco-c:
Another two observations:
1) Rectangle areas might be limiting, but it's fine for now until we actually find a case where rectangles are not enough;
2) The UI is not really that usable. When you press 'm' new windows get opened with the images. It would be better if: 1) we only opened the images once, not twice; 2) we opened the images in a single window side by side, like it is now.

sagarvijaygupta:
@marco-c  I will make the required control shift

For the two observations:
1. If we will feel the need I will try to add `linestrings` for polygon selection if that could suffice the need.
2. This will require complete removal of `tkinter` based interface. I hope you will be fine with that. Though it will be quite tough to show both images in one window because they will have to be concatenated in one single image with a boundary line in between and then further calculation for boxes will be taken in relatively to the concatenated image.

marco-c:
> If we will feel the need I will try to add linestrings for polygon selection if that could suffice the need.

Ok, we will think about it if it actually becomes a need.

> This will require complete removal of tkinter based interface. I hope you will be fine with that. Though it will be quite tough to show both images in one window because they will have to be concatenated in one single image with a boundary line in between and then further calculation for boxes will be taken in relatively to the concatenated image.

Yes, removing the tkinter based interface is fine. And it's fine to have multiple windows too, if a single window complicates things.
Or, is there a way to integrate opencv into tkinter?

sagarvijaygupta:
@marco-c  I did google about integrating opencv with tkinter but it doesn't provide exactly what we are looking for. I will first show you the interface with multiple windows (where I have fixed the location of windows so that they don't move). If still we feel the need to have it in one window I will find a solution to it.

marco-c:
> I will first show you the interface with multiple windows (where I have fixed the location of windows so that they don't move)

Sounds good, almost equivalent as having a single window

hardik124:
I have a tkinkter tool which can help us label the data, it is designed for
YOLO like networks, but with few modifications it should work fine.

marco-c:
> I have a tkinkter tool which can help us label the data, it is designed for YOLO like networks, but with few modifications it should work fine.

How does it work? Can you show it to us?

sagarvijaygupta:
@hardik124 are you talking about BBox-Label-Tool or is something else?

sagarvijaygupta:
@marco-c Regarding #126 I think I should add it in separate PR if you want. 

Now the labeling works the way you suggested.
> press y to mark compatible;
    press d to mark compatible but with differences, let the user select the regions and then press Enter when they are done;
    press n to mark incompatible, let the user select the regions and then press Enter when they are done.

Though labeling assigned on pressing `d` is `d` only and not `y`.

hardik124:
@marco-c  please check this repo https://github.com/puzzledqs/BBox-Label-Tool

hardik124:
@sagarvijaygupta  I am talking of BBoX tool.

autowebcompat/utils.py
```diff
@@ -195,3 +196,22 @@ def write_labels(labels, file_name='labels.csv'):
...
         for key, values in labels.items():
             writer.writerow([key, values])
+
+
+def read_boundary_boxes(file_name):
+    try:
+        with open(file_name, 'r') as f:
+            next(f)
+            reader = csv.reader(f)
+            boundary_boxes = {row[0]: ast.literal_eval(row[1]) for row in reader}
```
marco-c: Can we store the box with json? It should be more reliable

autowebcompat/utils.py
```diff
@@ -195,3 +196,22 @@ def write_labels(labels, file_name='labels.csv'):
         writer.writerow(["Image Name", "Label"])
         for key, values in labels.items():
             writer.writerow([key, values])
+
+
+def read_boundary_boxes(file_name):
```
marco-c: Let's always use the terminology `bounding box` instead of `boundary box`.

generate_labels.py
```diff
@@ -11,6 +11,8 @@
 ydn_reverse_map = {0: 'y', 1: 'd', 2: 'n'}
 
 for file_name in all_file_names:
+    if 'boundary_box' in file_name:
```
marco-c: We need to think of a way to manage this, the bounding boxes made by different people might differ.

label.py
```diff
@@ -11,80 +11,106 @@
...
-panel1.pack(side="left", padx=10)
-panel2 = Label(root)
-panel2.pack(side="left", padx=10)
-panel3 = Label(root)
-panel3.pack(side="left", padx=10)
+drawing = False
+key_map = {"Escape": 27, "r": 114, "Enter": 13, "y": 121, "n": 110, "d": 100}
+cv2.namedWindow('firefox')
+cv2.namedWindow('chrome')
+
```
marco-c: We should add the window which shows differences back, it was useful to perform labeling more quickly.

label.py
```diff
@@ -11,80 +11,106 @@
...
+    print("Reading %s" % current_image)
+    firefox_ss = cv2.imread("data/%s_firefox.png" % current_image)
+    chrome_ss = cv2.imread("data/%s_chrome.png" % current_image)
+    drawing_area_firefox = np.zeros((firefox_ss.shape), np.uint8)
+    drawing_area_chrome = np.zeros((chrome_ss.shape), np.uint8)
+    boxes_firefox = []
+    boxes_chrome = []
+    cv2.setMouseCallback('firefox', draw_boundary_box, [drawing_area_firefox, boxes_firefox])
+    cv2.setMouseCallback('chrome', draw_boundary_box, [drawing_area_chrome, boxes_chrome])
+    visiblity = 1
```
marco-c: typo here, it should be `visibility`

label.py
```diff
@@ -11,80 +11,106 @@
...
-def callback_y(e):
-    labels[current_image] = 'y'
-    get_new_image()
-
-
-# The image are basically the same, except for advertisement or content.
-def callback_d(e):
-    labels[current_image] = 'd'
+    print("Reading %s" % current_image)
+    firefox_ss = cv2.imread("data/%s_firefox.png" % current_image)
```
marco-c: Use a more descriptive name, `firefox_ss` is confusing

label.py
```diff
@@ -11,80 +11,106 @@
...
+    chrome_ss = cv2.imread("data/%s_chrome.png" % current_image)
+    drawing_area_firefox = np.zeros((firefox_ss.shape), np.uint8)
```
marco-c: Put this stuff in a `reset_bounding_boxes` function, so you don't have to repeat the same code here and when `r` is presseed

label.py
```diff
@@ -11,80 +11,106 @@
...
+        elif k == key_map["r"]:
+            drawing_area_firefox = np.zeros((firefox_ss.shape), np.uint8)
+            drawing_area_chrome = np.zeros((chrome_ss.shape), np.uint8)
+            boxes_chrome = []
+            boxes_firefox = []
+            cv2.setMouseCallback('firefox', draw_boundary_box, [drawing_area_firefox, boxes_firefox])
+            cv2.setMouseCallback('chrome', draw_boundary_box, [drawing_area_chrome, boxes_chrome])
+        # <Return> saves the current marking and moves to next image or skips it if no label is assigned
+        elif k == key_map["Enter"]:
+            if current_image not in labels.keys():
```
marco-c: When can this happen?

marco-c:
https://github.com/puzzledqs/BBox-Label-Tool would require changes to adapt it to our repo, so I think it's fine to write our own tool (given that it's simple enough). @hardik124 do you think there is something in that tool that the tool @sagarvijaygupta is writing doesn't support?

autowebcompat/utils.py
sagarvijaygupta: I will add it into json file. Should I print it in the pretty format with `indent=4`? To keep it readable?

label.py
sagarvijaygupta: Are you talking about the third window which was added with overlay among two screenshots?
>@marco-c Regarding #126 I think I should add it in separate PR if you want.
So I will add it in this PR only!

label.py
sagarvijaygupta: I renamed `ss` to `screenshot`.

label.py
sagarvijaygupta: Its for the same feature you added in `tkinter` based interface where if a person is unsure of the type and wants to skip marking that image .

sagarvijaygupta:
@marco-c  I will be adding `arrow` keys control on bounding boxes to shift them in 4 directions and also a cross button on each boundary box to remove one of them rather pressing `r`

autowebcompat/utils.py
marco-c: No, let's put it in a single line, it should be readable anyway since it's just an array of arrays.

label.py
marco-c: Yes, we shouldn't regress that functionality.

label.py
marco-c: Maybe we should have a different key for skipping, to avoid confusion (Enter -> Finish bounding box marking, Space -> Skip image).

label.py
sagarvijaygupta: I will add the functionality after #149 is resolved.

label.py
sagarvijaygupta: Done!

sagarvijaygupta:
@marco-c  I am getting `./label.py:2:1: I003 isort expected 1 blank line in imports, found 0` in flake8 and I am unable to resolve it. Could you please help?

Shashi456:
@sagarvijaygupta  you need to add a blank line before you import tkinter, that should solve it

sagarvijaygupta:
@Shashi456 Thanks!

sagarvijaygupta:
@marco-c  Please review this!

marco-c:
I've been thinking about this, I think there are two things we have to do:
1) When drawing a bounding box in one of the images, the same bounding box should be drawn on the other image automatically (as that's the location of the incompatibility in both images).
2) There might be images that both contain "fake" incompatibilities (type D) and "true" incompatibilities (type N). We should somehow let the user mark both types. Maybe press D to mark "fake" incompatibilities, press ENTER to finish, press N to mark "true" incompatibilities, press ENTER to finish. The bounding boxes for D and N should be colored differently. What do you think? If you can think of another solution, feel free to suggest.

sagarvijaygupta:
@marco-c I was also thinking about it.
1. For having same bounding box in both is definitely required but should we also add a label to it? Like whether it was drawn on firefox or chrome? I mean to say that presently it stores the bounding box coordinates in which window it was drawn (firefox or chrome). I think we should keep it the same way (and just draw a symmetrical box on the other side) because in future that will help us know which one is the right a part of an image and which one is the incompatibility corresponding to that.
2. For this I had though of two ideas. Keeping the control on right click and left click. The second idea was to have simple toggle button in the bounding box itself. It will feel simpler this way.

marco-c:
Agree about 1 and for 2 I like your second idea, it sounds easier for the user (and if they do a mistake, they can just click the button to change it).

sagarvijaygupta:
@marco-c I had some questions though:
1. For saving purposes should I add a sub-dict? Like `boundary_box['firefox']['n'] = []` or something else? 
2. Images should be marked 'n' only. Right? 
3. For images marked as 'd' toggle option should be disabled only?

marco-c:
1. Yes, sounds good.
2. What do you mean? If there are only 'd' inconsistencies, then the image should be marked 'd'.
3. Maybe we can do this: hit ENTER to enter regions, enter regions and choose the type with the toggle button (the type of the region being 'n' by default), hit ENTER to finish entering regions.

We could also stop storing the labels in a file and just store bounding boxes, then change read_labels in utils to generate the labels using the bounding box files.

sagarvijaygupta:
@marco-c I understood. We no longer have to press 'n' and 'd' rather toggle button does the job. I will make the necessary changes.

sagarvijaygupta:
@marco-c I made the necessary changes.
I haven't implemented the idea to remove `csv` files as after removal we won't be able to run `generate_labels.py` to create `labels.csv`. We will have to rename `read_labels` in `utils.py` one for `label.py` which will read from `NAME_bounding_box.json` and other for `train.py` which reads the final `labels.csv` until we decide how to create the `FINAL_bounding_box.json` which will be the some functions of all bounding boxes created by all users.

sagarvijaygupta:
@marco-c Please review it!

marco-c:
There are some conflicts to solve before I can merge this.

sagarvijaygupta:
@marco-c I changed `"` to `'` in the whole code. Please review it! Also travis build encountered some error.

marco-c:
I think the Travis failure is a network issue. I'll respin the build tomorrow.

PR #129, opened 2018-03-14
Cleaning up Data Inconsistencies.py

Shashi456:
After doing this update the number of inconsistencies got updated : 
```[*] Getting inconsistencies in screenshots...
[*] 5682 inconsistencies found.
[*] Writing to inconsistencies.csv... Done!
Number of photos: 16742
Number of pairs of images: 5530
Number of pairs of images possible: 11212
Percentage of Firefox inconsistencies: 0
Percentage of Chrome inconsistencies: 100 ```


Shashi456:
It's because some image files were being hidden due to the listdir function. for example in the image 40_0_firefox and 40_firefox only 40_0_firefox was being shown. I think as far as i have seen , all such inconsistent cases of xxxx_firefox.png and xxxx_0_firefox.png have now been cleared due to this update, As @sagarvijaygupta  pointed out this was because of the text files being read in the listdir function

marco-c:
Thanks!

PR #133, opened 2018-03-17
Add argparse interface to pretrain.py and train.py to select network architecture and optimizer

### LINKED ISSUE #79: Add argparse interface to pretrain.py and train.py to select network architecture and optimizer
Two steps here:
1) Add a function to network.py that returns the list of supported networks and a function that returns the list of supported optimizers;
2) Add [argparse](https://docs.python.org/3/library/argparse.html) interface to pretrain.py and train.py to select network and optimizer from those lists.

amit2rockon:
Fixes #79.

added function in network.py to return the list of supported network and optimizer.
added argparse interface to pretrain.py and train.py to select network architecture and optimizer

### DIFF
--- a/autowebcompat/network.py
+++ b/autowebcompat/network.py
@@ -120,7 +120,7 @@ def create_inception_network(input_shape):
 
 
 def create(input_shape, network='vgglike', weights=None):
-    assert network in SUPPORTED_NETWORKS , '%s is an invalid network' % network
+    assert network in SUPPORTED_NETWORKS, '%s is an invalid network' % network
     network_func = globals()['create_%s_network' % network]
     base_network = network_func(input_shape)
 
--- a/autowebcompat/network.py
+++ b/autowebcompat/network.py
@@ -116,7 +116,7 @@ def create_inception_network(input_shape):
     return Model(input, x)
 
 
-def create(input_shape, network='vgglike', weights=None):
+def create(input_shape, network, weights=None):
     network_func = globals()['create_%s_network' % network]
     base_network = network_func(input_shape)
 
@@ -155,7 +155,7 @@ def accuracy(y_true, y_pred):
     return K.mean(K.equal(y_true, K.cast(y_pred < 0.5, y_true.dtype)))
 
 
-def compile(model, optimizer='sgd', loss_func=contrastive_loss):
+def compile(model, optimizer, loss_func=contrastive_loss):
     allOptimizers = {
         'sgd': SGD(lr=0.0003, decay=1e-6, momentum=0.9, nesterov=True),
         'adam': Adam(),
@@ -166,3 +166,13 @@ def compile(model, optimizer='sgd', loss_func=contrastive_loss):
     opt = allOptimizers[optimizer]
 
     model.compile(loss=loss_func, optimizer=opt, metrics=[accuracy])
+
+
+def supported_network():
+    supp_net = ['inception', 'vgglike', 'vgg16']
+    return supp_net
+
+
+def supported_optimizer():
+    supp_opt = ['sgd', 'adam', 'nadam', 'rms']
+    return supp_opt
--- a/autowebcompat/network.py
+++ b/autowebcompat/network.py
@@ -120,6 +120,7 @@ def create_inception_network(input_shape):
 
 
 def create(input_shape, network='vgglike', weights=None):
+    assert network in SUPPORTED_NETWORKS , '%s is an invalid network' % network
     network_func = globals()['create_%s_network' % network]
     base_network = network_func(input_shape)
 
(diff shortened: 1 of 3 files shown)


amit2rockon:
@marco-c  please review this.

autowebcompat/network.py
```diff
@@ -166,3 +166,13 @@ def compile(model, optimizer='sgd', loss_func=contrastive_loss):
     opt = allOptimizers[optimizer]
 
     model.compile(loss=loss_func, optimizer=opt, metrics=[accuracy])
+
+
+def supported_network():
```
marco-c: Define these at the beginning of the file. Also, they don't need to be functions but can be lists directly:
```Python
SUPPORTED_NETWORKS = ['inception', 'vgglike', 'vgg16']
```

autowebcompat/network.py
```diff
@@ -116,7 +116,7 @@ def create_inception_network(input_shape):
     return Model(input, x)
 
 
-def create(input_shape, network='vgglike', weights=None):
+def create(input_shape, network, weights=None):
     network_func = globals()['create_%s_network' % network]
```
marco-c: Add an assertion at the beginning that the network is in the list of supported networks. Let's leave the default as it was.

autowebcompat/network.py
```diff
@@ -155,7 +155,7 @@ def accuracy(y_true, y_pred):
     return K.mean(K.equal(y_true, K.cast(y_pred < 0.5, y_true.dtype)))
 
 
-def compile(model, optimizer='sgd', loss_func=contrastive_loss):
+def compile(model, optimizer, loss_func=contrastive_loss):
```
marco-c: Add an assertion at the beginning that the optimizer is in the list of supported optimizers. Let's leave the default as it was.

pretrain.py
```diff
@@ -5,6 +5,15 @@
 
 from autowebcompat import network
 from autowebcompat import utils
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("network_", type=str, choices=network.supported_network(), help="Select the supported network")
```
marco-c: Why the final "_"?

marco-c: Change the help to "Select the network to use for training".

pretrain.py
```diff
@@ -5,6 +5,15 @@
...
+parser.add_argument("optimizer_", type=str, choices=network.supported_optimizer(), help="select the supported optimizer")
```
marco-c: Change the help to "Select the optimizer to use for training".

pretrain.py
```diff
@@ -5,6 +5,15 @@
...
+args = parser.parse_args()
+
+supp_net = args.network_
```
marco-c: No need to define these variables, use directly `args.network` when needed.

train.py
```diff
@@ -2,7 +2,15 @@
```
marco-c: Same comments apply to this file.

autowebcompat/network.py
amit2rockon: sir, this line in train.py or pretrain.py
```parser.add_argument("network", type = str, choices = SUPPORTED_NETWORKS, help = "Select the supported network")```
 already "choices"  checks if the given input 'network' is in the list or not 
if it is not present then it throws error 
``` pretrain.py: error: argument network: invalid choice: 'maxnet' (choose from 'inception', 'vgglike', 'vgg16')``` where maxnet is the given argument.
do we still need to add assertion to it ?

File: autowebcompat/network.py (hunk 3 above)
Comment (amit2rockon) [reply autowebcompat/network.py]: same with this case also

File: autowebcompat/network.py (hunk 2 above)
Comment (marco-c) [reply autowebcompat/network.py]: Yes, because this module could be in the future used by other scripts too, so it should not depend on what the callers do.

Comment (amit2rockon) [reply autowebcompat/network.py]: Added assertion in network.py

File: pretrain.py [hunk 8]
```diff
@@ -3,7 +3,14 @@
 import random
 from urllib.parse import urlparse
 
-from autowebcompat import network, utils
+from autowebcompat.network import *
```
marco-c: I'd prefer the more explicit option (`from autowebcompat import network, utils`), where you know where the functions and variables are coming from.

train.py
```diff
@@ -1,6 +1,13 @@
 import random
 
-from autowebcompat import network, utils
+from autowebcompat.network import *
+from autowebcompat import utils
+import argparse
+
+parser = argparse.ArgumentParser()
+parser.add_argument("network", type=str, choices=SUPPORTED_NETWORKS, help="Select the supported network")
```
marco-c: Change the help text like in pretrain.py

autowebcompat/network.py
```diff
@@ -3,6 +3,9 @@
 from keras.models import Model
 from keras.optimizers import SGD, Adam, Nadam, RMSprop
 
+SUPPORTED_NETWORKS = ['inception', 'vgglike', 'vgg16']
+SUPPORTED_OPTIMIZERS = ['sgd', 'adam', 'nadam', 'rms']
```
marco-c: We should merge this and `allOptimizers` from `compile` into a single variable, but we can do it in a follow-up.

marco-c:
@amit2rockon there's still a Travis failure.

PR #139, opened 2018-03-17
Remove images for which labels are not present 

### LINKED ISSUE #138: Training script should ignore images for which we don't have labels

Shashi456:
Fixes #138.

``` python
for i in images:
        if i not in img_labels:
           img.remove(i)
````
is having some unexpected behaviour so i avoided that . As in i still have images which have no labels present , so the error persists 
So i tried asserting the existence of the file in both labels and all images and this works and i have run it in my local repo so i have no other label issues anymore.

Another thing i observed is,  We could just load the images from labels . 

One last thing is labels.csv has an error , it has a tuple named "3516_rfc.index" i think we should remove that.

### CODE REVIEW

File: train.py [hunk 1]
```diff
@@ -1,18 +1,33 @@
+import csv
 import random
 
 from autowebcompat import network, utils
 
 labels = utils.read_labels()
 
+def update_all_image_names(images):
+    with open('labels.csv') as csvfile:
```
marco-c: You are re-reading the file that has already been read with utils.read_labels()  just a few lines earlier.

train.py
```diff
@@ -1,18 +1,33 @@
...
+            img_labels.append(row[0])
+    img = []
+    for i in images:
+        if i in img_labels:
+           img.append(i)
+    return(img)
+
+
 utils.prepare_images()
 all_image_names = utils.get_images()
```
marco-c: Filtering with a list comprehension here seems cleaner than your current implementation.

train.py
```diff
@@ -1,18 +1,33 @@
...
 all_images = sum([[i + '_firefox.png', i + '_chrome.png'] for i in all_image_names], [])
+all_image_names = update_all_image_names(all_image_names)
+print(all_image_names)
 image = utils.load_image(all_images[0])
 input_shape = image.shape
 BATCH_SIZE = 32
 EPOCHS = 50
 
-
```
marco-c: Remove these unrelated newline changes.

train.py
```diff
@@ -1,12 +1,17 @@
...
 
 from autowebcompat import network, utils
 
 labels = utils.read_labels()
 
+def update_all_image_names(images):
+    return [i for i in images if i in labels]
+
 utils.prepare_images()
 all_image_names = utils.get_images()
```
marco-c: I think the filtering should be done here, otherwise `all_images` will contain images that are not actually used.

train.py
```diff
@@ -1,12 +1,17 @@
+import csv
 import random
 
 from autowebcompat import network, utils
 
 labels = utils.read_labels()
 
+def update_all_image_names(images):
```
marco-c: No need to define a function for this very simple code, just do the filtering inline

Shashi456:
@marco-c  does this work ?

marco-c:
> @marco-c does this work ?

Yes, much cleaner :)

Shashi456:
@marco-c  should i file a seperate issue for the labels.csv errors? and another thing i'm still getting errors in train.py namely the one below .

``` python 
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\utils\data_utils.py", line 630, in data_generator_task
    generator_output = next(self._generator)
  File "G:\Code\autowebcompat\autowebcompat\utils.py", line 114, in __next__
    return self.next()
  File "G:\Code\autowebcompat\autowebcompat\utils.py", line 128, in next
    y_batch[i] = label
ValueError: could not convert string to float: 'd'

Traceback (most recent call last):
  File ".\train.py", line 47, in <module>
    model.fit_generator(train_iterator, steps_per_epoch=train_couples_len / BATCH_SIZE, epochs=EPOCHS)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\legacy\interfaces.py", line 87, in wrapper
    return func(*args, **kwargs)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\engine\training.py", line 2083, in fit_generator
    generator_output = next(output_generator)
StopIteration
```

marco-c:
Yes, please file an issue for the weird label error and another issue for this other problem you've just mentioned.

PR #141, opened 2018-03-17
Remove labels that don't correspond to any image

### LINKED ISSUE #140: Removing Deformed labels from the labels.csv
Labels.csv has some file names which are not present in images , probably caused due to some error while writing .
Namely : 
3516_rfc.index
10108_a-albumset-push.0

Shashi456:
Fixes #140.


marco-c:
They should be removed from the other labels file too, `amit2rockon.csv` under `label_persons`.

PR #146, opened 2018-03-18
Removing one maxpooling layer in vgg16 architecture

amit2rockon:
Since our input size is 32x24x3
so in vgg16 input dimension is reduced to  -negative dimension (invalid) due to strides in the last max-pooling layer, so one max-pooling layer must be removed.
    # Block 1 output  size  16x12x64
    # Block 2 output  size  8x4x128
    # Block 3 output  size  4x2x256
    # Block 4 output  size  2x1x512
    # Block 5 output size  1x-1x512 (invalid)
getting this error 
ValueError: Negative dimension size caused by subtracting 2 from 1 for 'max_pooling2d_5/MaxPool' (op: 'MaxPool') with input shapes: [?,2,1,512].

### DIFF
--- a/autowebcompat/network.py
+++ b/autowebcompat/network.py
@@ -39,7 +39,7 @@ def create_vgg16_network(input_shape):
     # Block 2
     x = Conv2D(128, (3, 3), activation='relu', padding='same')(x)
     x = Conv2D(128, (3, 3), activation='relu', padding='same',)(x)
-    x = MaxPooling2D(pool_size=(2, 2), strides=(2, 2))(x)
+    # x = MaxPooling2D(pool_size=(2, 2), strides=(2, 2))(x)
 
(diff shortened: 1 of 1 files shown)


Shashi456:
@amit2rockon but since we are using vgg16, shouldnt we stay true to the architecture and maybe increase dimensions size instead while using vgg16?

amit2rockon:
yes we can  increase the input size. and also inserting a  pooling layers there's a good chance we will see a decrease in the performance of the network (although it will train faster).Since max pool layers throw away information from the previous layers. In this case i think input size should be increase since we don't have large no of maxpool layers. @marco-c  please review this.

marco-c:
We shouldn't change the vgg16 default network, as we will likely use a pretrained one at some point.

PR #147, opened 2018-03-18
Adding Tests for checking if all the files for labels exists

### LINKED ISSUE #144: Add a test that checks that all files mentioned in labels files actually exist

Shashi456:
Fixes #144.

I was confused about where to add the test so i made a file. If it needs to be shifted i will do the same. 
@marco-c


tests/test_labels.py
```diff
@@ -0,0 +1,14 @@
+import csv
+import os
+
+
+def test_labels():
+    with open('..\labels.csv') as csvfile:
```
marco-c: There's a function in utils to read labels, why do you keep rewriting the reading manually? :smile:

tests/test_labels.py
```diff
@@ -0,0 +1,14 @@
...
+        reader = csv.reader(csvfile, delimiter=',')
+        for row in reader:
+            if row[0] == 'Image Name':
+                continue
+            file_name_firefox = '..\data\\' + row[0] + '_' + 'firefox.png'
```
marco-c: Nit: Use os.path.join

tests/test_labels.py
Shashi456: I forget about the functions :P

tests/test_labels.py
```diff
@@ -0,0 +1,12 @@
+import os
+
+from autowebcompat import utils
+
+
+def test_labels():
+    labels = utils.read_labels()
+    for label in labels:
+        if label == 'Image Name':
```
marco-c: The read_labels function skipts the header, so this should be unnecessary.

tests/test_labels.py
```diff
@@ -0,0 +1,12 @@
...
+            continue
+        for browser in ['firefox', 'chrome']:
+            assert os.path.join('..\data\\', '{}_{}.png'.format(label, browser))
```
marco-c: With os.path.join there's no need to use the `\` characters.

PR #148, opened 2018-03-18
Store DOM tree alongside screenshots

### LINKED ISSUE #117: Store current page source and DOM tree alongside screenshots
We might use this data in the future to improve the classification.

vibss2397:
Fixes #117.

hi i created the function which if run will keep on storing the Dom tree in a list form and the source code, @marco-c  do take a look and tell if any changes are required

I had prior engagements and was busy for the past few days, and so couldn't work on it early, will continue contributing regularly now onwards


vibss2397:
@marco-c please review

collect.py
```diff
@@ -202,7 +202,19 @@ def screenshot(driver, file_path):
     image = Image.open(file_path)
     image.save(file_path)
 
-
+    
+def get_code(driver, file_path):
```
marco-c: Split this in two functions, one to store the source of the page and one to store the DOM tree.

collect.py
```diff
@@ -214,6 +226,7 @@ def run_test(bug, browser, driver, op_sequence=None):
         print('Continuing...')
 
     screenshot(driver, 'data/%d_%s.png' % (bug['id'], browser))
```
marco-c: Let's add a new function that calls the functions to take a screenshot, store the page source and store the DOM tree.

collect.py
```diff
@@ -306,6 +320,8 @@ def main(bugs):
 
 if __name__ == '__main__':
     random.shuffle(bugs)
+    main(bugs)
```
marco-c: This should be removed. We are already calling the function using multiple threads below.

collect.py
```diff
@@ -202,7 +202,19 @@ def screenshot(driver, file_path):
     image = Image.open(file_path)
     image.save(file_path)
 
-
+    
+def get_code(driver, file_path):
+    source = driver.page_source
+    f = open(file_path, "w", encoding="utf-8")
+    domtree = driver.find_elements_by_css_selector("*")
```
marco-c: Your code is only storing the tag names, it should store the entire DOM model.

marco-c: Maybe we could store the page source only and then parse it later to get the DOM tree.

marco-c: That is, we could store `document.documentElement.outerHTML` only.

vibss2397: isn't this how a dom model is supposed to be like

marco-c: Your code is only storing the tag names of the elements and is ignoring everything else. We have to store everything instead.

You can store `document.documentElement.outerHTML`, which will contain the entire html of the page and then we can parse it using a html library.

vibss2397:
@marco-c check out

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
     image.save(file_path)
 
 
+def get_code(driver, file_path):
+    source = driver.page_source
```
marco-c: Is this equivalent to `document.documentElement.outerHTML`?

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
...
+    f = open(file_path, "w", encoding="utf-8")
```
marco-c: Use a context manager here, don't close the file manually.

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
...
+    f.write("source code : " + source + "\n")
+    f.close()
+
+
+def do_combined_task(driver, location):
```
marco-c: The name needs to be improved, `do_combined_task` is too generic.

collect.py
vibss2397: yes it does

marco-c: Are you 100% sure? Can you show me the documentation where you found that they are equivalent?

collect.py
```diff
@@ -203,6 +203,19 @@ def screenshot(driver, file_path):
     image.save(file_path)
 
 
+def get_code(driver, file_path):
+    if(not(os.path.isdir('source'))):
```
marco-c: No need to check if the directory exists, **utils.mkdir** works both if the directory already exists and if it doesn't.

collect.py
```diff
@@ -203,6 +203,19 @@ def screenshot(driver, file_path):
...
+        utils.mkdir('source')
+    source = driver.page_source
+    with open(file_path, "w", encoding="utf-8") as open_file:
+        open_file.append("source code : " + source + "\n")
```
marco-c: Don't append, just write.
Also don't write the "source code :" and the newline at the end, they are unnecessary.

Nit: rename open_file to f
Nit: no need to store the **driver.page_source** in a variable if you are using it only once

collect.py
```diff
@@ -203,6 +203,19 @@ def screenshot(driver, file_path):
...
+
+
+def get_screenshot_code(driver, location):
```
marco-c: **get_screenshot_code** is not a great name, it can be confusing.
Use **get_screenshot_and_domtree**.

collect.py
```diff
@@ -203,6 +203,19 @@ def screenshot(driver, file_path):
...
+    screenshot(driver, 'data/' + location + '.png')
+    get_code(driver, 'source/' + location + '.txt')
```
marco-c: Store everything in the same directory **data**. The name of the file containing the DOM tree can be called dom_LOCATION

Also change the **location** name to something more meaningful.

collect.py
vibss2397: here are a few websites including the official selenium website, that dont technically compare them both but states the function of this method-

1. https://seleniumhq.github.io/selenium/docs/api/py/webdriver_remote/selenium.webdriver.remote.webdriver.html?highlight=page_source#selenium.webdriver.remote.webdriver.WebDriver.page_source
2.  https://stackoverflow.com/questions/7861775/python-selenium-accessing-html-source

i also ran it on my local machine and here are the results-

marco-c: So, document.documentElement.outerHTML doesn't get the source of the page, but the DOM tree.
The documentation you shared says "page_source" returns the source of the page.
As soon as you open a page, like you did in your example, they might be equivalent. But what if some JavaScript script on the page changes something in the DOM, or if the user writes something in a text field, etc.? Are they still the same in this case or does page_source return just the original source of the page?

vibss2397: @marco-c  you're right, my bad, the page source wont change with javascript
https://stackoverflow.com/questions/17333582/selenium-do-javascript-actions-reflect-in-driver-page-source

i fixed it, check it

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
...
+def get_code(driver, file_path):
+    wait_loaded(driver)
+
+    with open(file_path, 'w', encoding='utf-8') as f:
+        f.write(driver.execute_script('return document.documentElement.outerHTML'))
+
+
+def get_screenshot_and_domtree(driver, file_name):
+    screenshot(driver, 'data/' + file_name + '.png')
+    get_code(driver, 'data/' + 'dom_LOCATION_' + file_name + '.txt')
```
marco-c: Don't actually put LOCATION here, it was just a placeholder to show you what the name should be 😄

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
     image.save(file_path)
 
 
+def get_code(driver, file_path):
+    wait_loaded(driver)
```
marco-c: Since both this function and screenshot are using **wait_loaded**, let's move it in **get_screenshot_and_domtree**. We don't need to wait for the page to be loaded twice.

collect.py
```diff
@@ -203,6 +203,18 @@ def screenshot(driver, file_path):
     image.save(file_path)
 
 
+def get_code(driver, file_path):
```
marco-c: The name of the function should be **get_domtree**

vibss2397:
@marco-c hopefully this should cover all the changes you requested

marco-c:
There are some errors on Travis.

vibss2397:
done with them

PR #152, opened 2018-03-20
Issue #151: add flake8-quotes

### LINKED ISSUE #151: Add flake8-quotes

rhcu:
Fixes #151.


rhcu:
Do we need to use single instead of double quotes?

marco-c:
> Do we need to use single instead of double quotes?

Yes.

PR #153, opened 2018-03-20
Changing Strings to Categories for Keras

### LINKED ISSUE #142: Convert string labels to categories
The following error comes up when i try to run train.py

``` python 
Exception in thread Thread-1:
Traceback (most recent call last):
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\threading.py", line 916, in _bootstrap_inner
    self.run()
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\threading.py", line 864, in run
    self._target(*self._args, **self._kwargs)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\utils\data_utils.py", line 630, in data_generator_task
    generator_output = next(self._generator)
  File "G:\Code\autowebcompat\autowebcompat\utils.py", line 114, in __next__
    return self.next()
  File "G:\Code\autowebcompat\autowebcompat\utils.py", line 128, in next
    y_batch[i] = label
ValueError: could not convert string to float: 'd'

Traceback (most recent call last):
  File ".\train.py", line 47, in <module>
    model.fit_generator(train_iterator, steps_per_epoch=train_couples_len / BATCH_SIZE, epochs=EPOCHS)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\legacy\interfaces.py", line 87, in wrapper
    return func(*args, **kwargs)
  File "C:\Users\pawan sasanka\AppData\Local\Programs\Python\Python36\lib\site-packages\keras\engine\training.py", line 2083, in fit_generator
    generator_output = next(output_generator)
StopIteration
```

Shashi456:
So using read labels i accessed the dictionary of the labels , and then according to the classifier type im changing the labels to 0 or 1 . I'm currently returning a dictionary for the same. 
Fixes #142. 

Oh looks like i forgot to return the labels dict , i will do that in the next commit .
@marco-c please review this.


marco-c:
This doesn't fix #142, as you are just adding a function in utils.py, without actually changing anything in the training script.
Please also change train.py to actually use the correct labels, otherwise this doesn't help much.

I think we don't need a dictionary, it's just a waste of memory. We can just compute the label dinamically similar to what #154 is proposing.

autowebcompat/utils.py
```diff
@@ -179,12 +179,16 @@ def make_infinite(gen_func, elems):
         yield from gen_func(elems)
 
 
-def read_labels(file_name='labels.csv'):
+def read_labels(encoding, file_name='labels.csv'):
```
marco-c: Let's use 'Y vs D+N' as the default here, and let's replace **encoding** with something like **classification_type**

Shashi456:
@marco-c  are the test cases breaking because i've written a dictionary comprehension?

marco-c:
@Shashi456 have you seen the Travis logs?

Shashi456:
@marco-c  yes it's said that the read labels function requires an additional argument in test_utils.py , makes sense. i was afraid that it would break the test cases .

Shashi456:
@marco-c  does this work?  I have made a simple function to do the same, similar to my original approach.

Shashi456:
@marco-c  can you take a look at this

marco-c:
This is partly the same as the first iteration of the patch, so the same comment applies:
> I think we don't need a dictionary, it's just a waste of memory. We can just compute the label dinamically similar to what #154 is proposing.

Shashi456:
@marco-c what other way could I look at? Because #154 directly does it in the function itself. But that would probably break other files in which we are using read_labels . Should I directly write a function which does it at run time ?

marco-c:
> @marco-c what other way could I look at? Because #154 directly does it in the function itself. But that would probably break other files in which we are using read_labels . Should I directly write a function which does it at run time ?

Yes, similar to #154, not exactly the same :)
A function that does it at runtime seems the best option.

vrishank97:
@Sashi456 you can change the default argument for the encoder in #154 to none, and have it output unaltered labels

autowebcompat/utils.py
```diff
@@ -179,12 +179,17 @@ def make_infinite(gen_func, elems):
         yield from gen_func(elems)
 
 
-def read_labels(file_name='labels.csv'):
+def read_labels(file_name='labels.csv', classification_type = None):
```
marco-c: I prefer a separate function to handle this, rather than overloading read_labels with additional meanings.

Shashi456: @marco-c  what if instead of writing a seperate function we directly modified content using 

``` python
            if classification_type == 'Y vs D + N':
                labels = {x: 1 if y == 'y' else 0 for x,y in labels.items}
            elif classification_type == 'Y + D vs N':
                labels = {x: 0 if y == 'n' else 1 for x,y in labels.items}
```

or I could pass the dictionary to a simple function and overwrite the labels, instead of making a new dictionary and returning it like I did earlier.

marco-c: > @marco-c what if instead of writing a seperate function we directly modified content using

Better as a function, so you can easily write some test for it.

> I could pass the dictionary to a simple function and overwrite the labels, instead of making a new dictionary and returning it like I did earlier.

It seems simpler to me to just have a function that, given a label, returns the categorical value.

marco-c:
Yes, this looks better!
Please make sure to write some tests for the new function you have added.

autowebcompat/utils.py
```diff
@@ -190,6 +190,19 @@ def read_labels(file_name='labels.csv'):
     return labels
 
 
+def Classify(label, classification_type='Y vs D + N'):
```
marco-c: **Classify** is not a great name for the function, use something like **to_categorical_label**.

Shashi456:
> Please make sure to write some tests for the new function you have added.

@marco-c  should I open an issue for the same or do I add them in the same PR?

marco-c:
> @marco-c should I open an issue for the same or do I add them in the same PR?

Add in the same PR.

Shashi456:
@marco-c  i was wondering if we should add a test in test_labels.py to also check if all the labels belong to y or d or n and also that no id goes label-less

marco-c:
@marco-c i was wondering if we should add a test in test_labels.py to also check if all the labels belong to y or d or n and also that no id goes label-less

The first, yes. The second, we can't do yet as we have many elements that haven't been labeled yet.
Can you open an issue for the first?

PR #161, opened 2018-03-24
Fix random seed

### LINKED ISSUE #160: Fix random seed
We need to fix the seed for `random` so that the test-train split doesn't vary and accuracy becomes consistent.

vrishank97:
Fixes #160


train.py
```diff
@@ -17,6 +17,7 @@
 input_shape = image.shape
 BATCH_SIZE = 32
 EPOCHS = 50
+random.seed = 100
```
marco-c: We should use the random.seed function.

PR #162, opened 2018-03-27
Add validation set in train and pretrain files

### LINKED ISSUE #150: Add validation set and callbacks in pretrain and train scripts
Currently, we only have train accuracy and its not being validated. We can go for a simple validation script or a K fold validation. 
With validation check , we can add callbacks such as early stopping and model checkpoints.

osmarcedron:
Adds the validation_set

Fixes #150.

### DIFF
--- a/pretrain.py
+++ b/pretrain.py
@@ -23,8 +23,10 @@ all_images = utils.get_all_images()[:SAMPLE_SIZE]
 image = utils.load_image(all_images[0])
 input_shape = image.shape
 
-TRAIN_SAMPLE = 90 * (SAMPLE_SIZE // 100)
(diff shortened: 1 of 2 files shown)


pretrain.py
```diff
@@ -6,6 +6,11 @@
 
 from autowebcompat import network, utils
 
+SAMPLE_SIZE = 3000
```
marco-c: While I like these changes, moving the constants at the top, they are unrelated to the objective of this PR, so they should be in a separate PR.

osmarcedron:
@marco-c, the changes with respect to the unrelated changes (#167) were addressed, could you please check this PR?

osmarcedron:
Done! @marco-c, check it again please.

PR #163, opened 2018-03-27
Test for get_ImageDataGenerator

### LINKED ISSUE #71: Write unit test for get_ImageDataGenerator function in utils.py

osmarcedron:
I am not very familiar with generators, anyone has a better idea about what to test?

Fixes #71.

### DIFF
--- a/tests/test_utils.py
+++ b/tests/test_utils.py
@@ -1,6 +1,8 @@
 import os
 
 from PIL import Image
+import keras
+from keras.preprocessing.image import ImageDataGenerator, img_to_array
 import numpy as np
 import pytest
 
@@ -53,6 +55,22 @@ def test_write_labels(tmpdir):
     utils.write_labels(label, file_name=file_path)
     assert(os.path.exists(file_path))
     assert(label == utils.read_labels(file_name=file_path))
+
+
+def test_get_ImageDataGenerator(tmpdir):
+    file_path = tmpdir.join('Image2.jpg')
+    img = Image.new('RGB', (30, 30), color=(255, 0, 0))
+    img.save(file_path.strpath)
+
+    data_gen = utils.get_ImageDataGenerator(['Image2.jpg'], (32, 24, 3), file_path.dirname)
+    assert(isinstance(data_gen, ImageDataGenerator))
+
+    x = img_to_array(img, data_format=keras.backend.image_data_format())
+    x = x.reshape((1,) + x.shape)
+    X_batch = data_gen.flow(x).next()
+    assert X_batch.shape[0] == x.shape[0]
+    assert X_batch.shape[1:] == x.shape[1:]
+    assert (X_batch >= 0).all() and (X_batch <= 1).all()
 
 
 test_balance_data = [
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -86,13 +86,13 @@ def load_image(fname, parent_dir='data_resized'):
     return x
 
 
-def get_ImageDataGenerator(images, image_shape):
+def get_ImageDataGenerator(images, image_shape, parent_dir='data_resized'):
     data_gen = ImageDataGenerator(rescale=1. / 255)
 
     x = np.zeros((len(images),) + image_shape, dtype=keras.backend.floatx())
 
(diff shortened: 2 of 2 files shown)


tests/test_utils.py
```diff
@@ -54,6 +56,13 @@ def test_write_labels(tmpdir):
     assert(os.path.exists(file_path))
 
 
+def test_get_ImageDataGenerator(tmpdir):
+    all_images = utils.get_all_images()
+    image = random.sample(all_images, 1)
+    data_gen = utils.get_ImageDataGenerator(image, (32, 24, 3))
+    assert(isinstance(data_gen, ImageDataGenerator))
+
```
marco-c: You should test that you can get an element from the iterator (and only one) and that the element you get is an image of the shape you selected.

tests/test_utils.py
```diff
@@ -54,6 +56,13 @@ def test_write_labels(tmpdir):
     assert(os.path.exists(file_path))
 
 
+def test_get_ImageDataGenerator(tmpdir):
+    all_images = utils.get_all_images()
+    image = random.sample(all_images, 1)
```
marco-c: It isn't great to use `random` in tests. Just use the first image here.

marco-c: And rename the variable. It's an array of images, so calling it `image` can be confusing.

osmarcedron:
@marco-c , I have a doubt about the implementation of get_ImageDataGenerator, according to the documentation the method fit() is ```only required if featurewise_center or featurewise_std_normalization or zca_whitening```. Why is it used if the generator only rescales the data?

marco-c:
> Why is it used if the generator only rescales the data?

Because we haven't decided yet about the transformations we will apply. And moreover using it even when you don't use featurewise_center or featurewise_std_normalization or zca_whitening doesn't cause any problem.

osmarcedron:
It was just curiosity... @marco-c, could you check it please.

marco-c:
> It was just curiosity...

Oh, ok, I thought it was blocking you somehow!

tests/test_utils.py
```diff
@@ -55,6 +56,19 @@ def test_write_labels(tmpdir):
     assert(label == utils.read_labels(file_name=file_path))
 
 
+def test_get_ImageDataGenerator():
+    img_name = next(f for f in os.listdir('data/') if f.endswith('.png'))
+    img = utils.load_image(img_name)
```
marco-c: Let's create an image in memory here, rather than loading a real image from the data directory (see the `test_load_image` test for an example).

tests/test_utils.py
```diff
@@ -55,6 +56,19 @@ def test_write_labels(tmpdir):
...
+    data_gen = utils.get_ImageDataGenerator([img_name], img.shape)
+    assert(isinstance(data_gen, ImageDataGenerator))
+    assert data_gen.rescale == 1. / 255
```
marco-c: No need to test this, we should test that the rescaling happens, but not the property itself.

tests/test_utils.py
```diff
@@ -55,6 +56,19 @@ def test_write_labels(tmpdir):
...
+
+    x = img.reshape((1,) + img.shape)
+    X_batch = data_gen.flow(x).next()
+    assert X_batch.shape[0] == 1
+    assert X_batch.shape[1:] == img.shape
+
```
marco-c: Here we should assert that the image is resized and that it is rescaled

tests/test_utils.py
```diff
@@ -55,6 +56,22 @@ def test_write_labels(tmpdir):
     assert(label == utils.read_labels(file_name=file_path))
 
 
+def test_get_ImageDataGenerator(tmpdir):
+    file_path = tmpdir.join('Image2.jpg')
+    img = Image.new('RGB', (30, 30), color=(255, 0, 0))
+    img.save(file_path.strpath)
+    img = utils.load_image('Image2.jpg', file_path.dirname)
```
marco-c: I think you don't need to reload the image.

osmarcedron:
@marco-c, check it please

marco-c:
Thanks!

PR #165, opened 2018-03-29
Expand the write_labels tests to make sure read_labels can read labels written by write_labels

### LINKED ISSUE #68: Expand the write_labels/read_labels tests to make sure read_labels can read labels written by write_labels
The test_write_labels test in the test_utils.py file just simply asserts the existence of the file, a new test should be written or the same updated. This test should write the labels and then read them back. The read labels should be the same as the written ones. 

See also PR #63.

anujanegi:
Fixes #68.


marco-c:
The test is failing, so you will have to change it a bit.
Make the `label` object more close to the actual one we are storing with labels.py (keys are strings, values are strings).

PR #168, opened 2018-04-05
[WIP] Analyse false predictions

Trion129:
TODO:
- [x] Load all images and make predictions
- [ ] Compare to extract false predictions
- [ ] Create a GUI to display false predicted images in sequence
- [ ] Be able to pop the image label if the images are too noisy to learn

### DIFF
--- a/analyse_false.py
+++ b/analyse_false.py
@@ -1,0 +1,60 @@
+from functools import lru_cache
+from urllib.parse import urlparse
+from keras.models import load_model
+from autowebcompat.network import contrastive_loss
+import itertools
+import numpy as np
+
+from autowebcompat import utils
+
+bugs = utils.get_bugs()
+
+utils.prepare_images()
+all_images = utils.get_all_images()[:3000]  # 3000
+image = utils.load_image(all_images[0])
+input_shape = image.shape
+BATCH_SIZE = 32
+EPOCHS = 50
+
+bugs_to_website = {}
+for bug in bugs:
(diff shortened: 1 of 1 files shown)


marco-c:
@Trion129 are you still interested in working on this?

marco-c:
ping @Trion129

Trion129:
Hi @marco-c, I was busy last month at college. I had made some progress, but I was having issue with placing the "next" button properly, it gets overlapped under the images. I'll push the code anyways, you might have a fix.

analyse_false.py
```diff
@@ -0,0 +1,112 @@
...
+    global panel1, panel2, panel1_text, panel2_text, image_index
+
+    truth, prediction, image_name = images_for_analysis[image_index]
+
+    firefox_name =  image_name + "_firefox.png"
+    chrome_name = image_name + "_chrome.png"
+
+    img1 = ImageTk.PhotoImage(Image.open("data/" + firefox_name))
+    panel1.configure(image=img1)
+    panel1.image = img1
```
Trion129: On adding the image, it covers the next button. Even though they are in separate grids.

Trion129:
Never-mind, it started to work when I put the button inside the image-view itself. Weird tkinter...

marco-c:
> @marco-c For the popping of images, as the pics are spread over many csv files, I think it is better to generate a new list "false_pairs.csv" where the image names will be appended. This list can be used by pretrain and train scripts to remove false pairs from data. What's your opinion?

I think we shouldn't pop those images, let's just implement the analysis of false predictions.

PR #169, opened 2018-04-06
Adding vgg19 to the network.py

Shashi456:
Adding vgg19 to the network.py


Shashi456:
@marco-c  so i will add resnet after, can you take a look at this right now?

PR #170, opened 2018-04-09
Update documentation to include argument parser

### LINKED ISSUE #166: Update Documentation to include argument Parser
We recently added argument parser for train and pretrain.py script i think we should update the documentation to reflect the same

Shashi456:
So i've included arguments in the documentation.
I was skeptical as to whether we should include the options for these arguments in the documentation or leave it to the user. 
I think the documentation should now look something like this. 
I will include the default value for these arguments.

Fixes #166.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
-### Training 
+### Training
 
 Now that we have a dataset with labels, we can train a neural network to automatically detect screenshots that are incompatible. We are currently using a [Siamese architecture](https://papers.nips.cc/paper/769-signature-verification-using-a-siamese-time-delay-neural-network.pdf) with different Convolutional Neural Networks, but are open to test other ideas.
 
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
-### Training
+### Training methodologies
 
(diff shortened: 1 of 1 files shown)


Shashi456:
@marco-c  please look at this

README.md
```diff
@@ -39,7 +39,7 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
-### Training
+### Training methodologies
```
marco-c: Don't change this, it's unrelated to the PR.

README.md
```diff
@@ -69,7 +69,21 @@ For the unsupervised training, we are using a related problem for which we alrea
 - Clone the repository with submodules: `git lfs clone --recurse-submodules REPO_URL`
 - Install the dependencies in requirements.txt: `pip install -r requirements.txt`.
 - Install the dependencies in test-requirements.txt: `pip install -r test-requirements.txt`.
-- Run the **pretrain.py / train.py** script to train the neural network.
+
+## Training the network
+- The **pretrain.py** or **train.py** script can be run to train the neural network.
+- There are options for running the same:
```
marco-c: Remove this line and add ", with the following options:" to the previous line.

README.md
```diff
@@ -69,7 +69,21 @@ For the unsupervised training, we are using a related problem for which we alrea
...
+
+    ```
+    -network                  For training different kinds of
```
marco-c: The description for this should be "To select which network architecture to use".

README.md
```diff
@@ -69,7 +69,21 @@ For the unsupervised training, we are using a related problem for which we alrea
...
+                              network
+                              
+    -optimizer                For training a network with different
```
marco-c: The description for this should be "To select the optimizer to use".

README.md
Shashi456: I thought "training" would be redundant and so i changed it . I'll revert it back

Shashi456:
@marco-c  i was wondering if we should include supported architectures and optimizers?

marco-c:
> @marco-c i was wondering if we should include supported architectures and optimizers?

No, I'm pretty sure if we do that they will get stale (as we will forget to update the README every time we add a new architecture or optimizer).

PR #171, opened 2018-04-10
Add early stopping and modelcheckpoint callbacks

### LINKED ISSUE #150: Add validation set and callbacks in pretrain and train scripts

osmarcedron:
Fixes #150.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,6 +79,8 @@ For the unsupervised training, we are using a related problem for which we alrea
     -optimizer                To select the optimizer to use   
                               
     -classification_type      Either Y vs N + D or Y + N vs D
+
+    --early_stoppping	      (Optional) To stop training when validation accuracy has stopped improving
     ```
 
 
--- a/pretrain.py
+++ b/pretrain.py
@@ -4,6 +4,8 @@ import itertools
 import random
 from urllib.parse import urlparse
 
+from keras.callbacks import EarlyStopping, ModelCheckpoint
+
 from autowebcompat import network, utils
 
 SAMPLE_SIZE = 3000
@@ -14,6 +16,7 @@ random.seed(42)
 parser = argparse.ArgumentParser()
 parser.add_argument('network', type=str, choices=network.SUPPORTED_NETWORKS, help='Select the network to use for training')
 parser.add_argument('optimizer', type=str, choices=network.SUPPORTED_OPTIMIZERS, help='Select the optimizer to use for training')
+parser.add_argument('--early_stopping', dest='early_stopping', action='store_true', help='Stop training training when validation accuracy has stopped improving.')
 args = parser.parse_args()
 
 bugs = utils.get_bugs()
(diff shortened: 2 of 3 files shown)


pretrain.py
```diff
@@ -74,11 +76,17 @@ def gen_func(images):
 model = network.create(input_shape, args.network)
 network.compile(model, args.optimizer)
 
-model.save('pretrain.h5')
+# serialize model to JSON
```
marco-c: Why did you change this?

pretrain.py
```diff
@@ -74,11 +76,17 @@ def gen_func(images):
...
+model_json = model.to_json()
+with open('model_pretrain.json', 'w') as json_file:
+    json_file.write(model_json)
 
-model.fit_generator(train_iterator, validation_data=validation_iterator, steps_per_epoch=train_couples_len / BATCH_SIZE, validation_steps=validation_couples_len / BATCH_SIZE, epochs=EPOCHS)
+filepath = 'best_pretrain_weights.hdf5'
```
marco-c: Remove this variable and use the value directly in the `ModelCheckpoint` constructor.

pretrain.py
```diff
@@ -74,11 +76,17 @@ def gen_func(images):
...
+early_stopping = EarlyStopping(monitor='val_accuracy', patience=2)
```
marco-c: Early stopping should be an option of the scripts.

train.py
```diff
@@ -1,6 +1,8 @@
 import argparse
```
marco-c: Same comments apply to this file.

pretrain.py
osmarcedron: As the ModelCheckpoint callback saves the full model, something similar to what model.save does, I thought the latter would be redundant.

marco-c: That sounds good, but why do we need to save the model as JSON? We probably can just remove the saving at the beginning altogether, no?

osmarcedron: @marco-c. Indeed, that was a mistake and it is already been removed, please check it now.

README.md
```diff
@@ -79,6 +79,8 @@ For the unsupervised training, we are using a related problem for which we alrea
     -optimizer                To select the optimizer to use   
                               
     -classification_type      Either Y vs N + D or Y + N vs D
+
+    --early_stoppping	      (Optional) To stop training when val_accuracy has stopped improving
```
marco-c: Let's use the full wording (`validation accuracy`) as it's clearer.

pretrain.py
```diff
@@ -14,6 +16,7 @@
 parser = argparse.ArgumentParser()
 parser.add_argument('network', type=str, choices=network.SUPPORTED_NETWORKS, help='Select the network to use for training')
 parser.add_argument('optimizer', type=str, choices=network.SUPPORTED_OPTIMIZERS, help='Select the optimizer to use for training')
+parser.add_argument('--early_stopping', dest='early_stopping', action='store_true', help='Stop training when a monitored quantity has stopped improving.')
```
marco-c: Here let's use the same text as the description in the README

marco-c: Without `To`

osmarcedron:
@marco-c, the comments were addressed. I think Travis is having problems trying to clone the submodules.

marco-c:
> I think Travis is having problems trying to clone the submodules.

Yep, I'll respin the build tomorrow. Hopefully the problem is transient.

PR #172, opened 2018-04-11
Hide scrollbar in chromedriver

### LINKED ISSUE #13: Make the screenshots in the different browsers equal in size
Right now, Chrome is showing a scrollbar that we should get rid of, as it is simply adding noise.
Once we remove the scrollbar, the screenshots are probably not the same size in Firefox and Chrome, so we should adjust it to be the same.

osmarcedron:
Fixes #13.

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -305,6 +305,7 @@ chrome_options = webdriver.ChromeOptions()
 chrome_options.binary_location = chrome_bin
 chrome_options.add_argument('--no-sandbox')
 chrome_options.add_argument('--headless')
+chrome_options.add_argument('--hide-scrollbars')
 chrome_options.add_argument('--window-size=412,732')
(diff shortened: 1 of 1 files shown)


marco-c:
Have you tested this? It looks like the flag has been removed: https://bugs.chromium.org/p/chromium/issues/detail?id=350136. Did they bring it back?

marco-c:
There's still no `--hide-scrollbars`: https://github.com/chromium/chromium/blob/master/content/public/common/content_switches.cc.

marco-c:
Oh, maybe it was brought back in headless mode: https://github.com/chromium/chromium/commit/d71fc504213b7d60f98d8b1db9abb6bda280a042.

marco-c:
Can you provide example screenshots before/after?

osmarcedron:
@marco-c sorry for not providing images (1st image before, 2nd image after).
5103_chrome
5103_chrome
I found the ```--hide-scrollbars``` flag [here](https://peter.sh/experiments/chromium-command-line-switches/)

PR #176, opened 2018-04-14
Update documentation with bounding-box labeling

### LINKED ISSUE #174: Update README.md with new labeling methodology

Shashi456:
Fixes #174

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,6 +39,18 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
+### Bounded-Box Labeling 
+
+This kind of labeling allows us to store the areas where the incompatibilities lie alongside the labels.
+<img src="https://user-images.githubusercontent.com/19645459/37787732-b21122b4-2e25-11e8-8d01-8d192023463f.png" />
+
+```
+Press 'y' to mark compatible.
+Press 'Enter' to select regions of incompatibility.
+Click 'T' to toggle on top left of boundary boxes.GREEN --> 'n' YELLOW --> 'd'
+Press 'Enter' to save changes
+``` 
+
 ### Training
 
 Now that we have a dataset with labels, we can train a neural network to automatically detect screenshots that are incompatible. We are currently using a [Siamese architecture](https://papers.nips.cc/paper/769-signature-verification-using-a-siamese-time-delay-neural-network.pdf) with different Convolutional Neural Networks, but are open to test other ideas.
--- a/README.md
+++ b/README.md
@@ -45,9 +45,9 @@ The bounding-box labeling allows us to store the areas where the incompatibiliti
 
 <img src="https://user-images.githubusercontent.com/18056781/39081665-10eda006-4563-11e8-9455-986b5a23934e.jpg" />
 
-- Press 'y' to mark the images as compatible.
-- Press 'Enter' to select the regions.
-- Click 'T' on top left of boundary boxes to toggle between colors. Green corresponds to 'n', yellow corresponds to 'd'.
+- Press 'y' to mark the images as compatible;
+- Press 'Enter' to select the regions;
+- Click the 'T' button in the top left corner of a boundary box to toggle between classes. Green corresponds to 'n', yellow corresponds to 'd';
 - Press 'Enter' to save changes. 
 
 ### Training
--- a/README.md
+++ b/README.md
@@ -39,17 +39,16 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
-### Bounded-Box Labeling 
+The Bounded-Box labeling allows us to store the areas where the incompatibilities lie.
 
-This kind of labeling allows us to store the areas where the incompatibilities lie alongside the labels.
-<img src="https://user-images.githubusercontent.com/19645459/37787732-b21122b4-2e25-11e8-8d01-8d192023463f.png" />
+<img src="https://user-images.githubusercontent.com/18056781/39081659-fdd4655e-4562-11e8-86f9-a5fab28634bf.JPG" />
 
-```
-Press 'y' to mark compatible.
-Press 'Enter' to select regions of incompatibility.
-Click 'T' to toggle on top left of boundary boxes.GREEN --> 'n' YELLOW --> 'd'
-Press 'Enter' to save changes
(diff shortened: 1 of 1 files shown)


Shashi456:
@marco-c  a couple of doubts, i took the image directly from #128  so i was wondering if i should upload cropped images individually and then add them to the readme. Is there anything else i should add for this labeling?

Another thing I've observed is the early stopping isn't really rendered properly in the readme file. should i fix that in this PR?

README.md
```diff
@@ -39,6 +39,18 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
+### Bounded-Box Labeling 
+
+This kind of labeling allows us to store the areas where the incompatibilities lie alongside the labels.
```
marco-c: This is the only available kind of labeling now

README.md
```diff
@@ -39,6 +39,18 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+<img src="https://user-images.githubusercontent.com/19645459/37787732-b21122b4-2e25-11e8-8d01-8d192023463f.png" />
+
+```
```
marco-c: Let's make this normal text and not a code block

README.md
```diff
@@ -39,6 +39,18 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+Press 'y' to mark compatible.
+Press 'Enter' to select regions of incompatibility.
+Click 'T' to toggle on top left of boundary boxes.GREEN --> 'n' YELLOW --> 'd'
```
marco-c: This sentence is a little unclear

README.md
```diff
@@ -39,6 +39,18 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
+### Bounded-Box Labeling 
+
+This kind of labeling allows us to store the areas where the incompatibilities lie alongside the labels.
+<img src="https://user-images.githubusercontent.com/19645459/37787732-b21122b4-2e25-11e8-8d01-8d192023463f.png" />
+
+```
+Press 'y' to mark compatible.
```
marco-c: Let's put all commands in list entries:
- Press **y** to mark the images as compatible;
- ...

sagarvijaygupta:
@Shashi456 @marco-c Should we not put images where we show which bounding box is marking `n` (not compatible) and which is marking `d` (compatible but different) where they actually exist and not randomly as in the present image?

marco-c:
> @Shashi456 @marco-c Should we not put images where we show which bounding box is marking n (not compatible) and which is marking d (compatible but different) where they actually exist and not randomly as in the present image?

Good idea, it might be better to make readers understand.
@Shashi456 see if you can find such examples.

marco-c:
> @sagarvijaygupta how do you actually run the bounded box labeling on a set of images ? I'd like to add that in the documentation as well.

Just run the label.py script

README.md
marco-c: Let's just remove the header

sagarvijaygupta:
@Shashi456 Run `python3 label.py file_name`

Shashi456:
@marco-c @sagarvijaygupta While trying to label these boxes, I thought it would be better if we made the labelling of the bounding boxes and labelling just the images as separate modules. I think that helps because bounded box labelling takes considerably more time than just looking at two photos and marking them as y,n or d. We could either make two separate label files or make them functions, and have a switch-case menu when label.py is run.

Also, I saw that in the same image multiple colours of bounding boxes are allowed. I think we should ensure that all boxes are of the same colour before storing them in the respective bounding box json.

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
+The Bounded-Box labeling allows us to store the areas where the incompatibilities lie.
```
marco-c: `The bounding-box labeling ...`

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+
+<img src="https://user-images.githubusercontent.com/18056781/39081659-fdd4655e-4562-11e8-86f9-a5fab28634bf.JPG" />
+
+<img src="https://user-images.githubusercontent.com/18056781/39081665-10eda006-4563-11e8-9455-986b5a23934e.jpg" />
+
+- Press 'y' to mark compatible.
```
marco-c: Press 'y' to mark the images as compatible.

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+- Press 'Enter' to select regions of incompatibility.
```
marco-c: to select the regions

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+- Click 'T' on top left of boundary boxes to toggle between colors. GREEN --> 'n' YELLOW --> 'd'
```
marco-c: on the top left corner of the bounding-boxes ... colors. Green corresponds to 'n', yellow corresponds to 'd'.

(It would also be nice to make `Green` actually green and `yellow` actually yellow)

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
...
+- Press 'Enter' to save changes
```
marco-c: `.` at the end.

README.md
Shashi456: @marco-c github doesn't allow colored text in markdown. It's an ongoing issue and as far as i have seen there is no way around it .

README.md
```diff
@@ -39,6 +39,17 @@ In the training phase, the best case is that we are able to detect between Y+D a
 
 The labeling technical details are described [in this issue](https://github.com/marco-c/autowebcompat/issues/2).
 
+The bounding-box labeling allows us to store the areas where the incompatibilities lie.
+
+<img src="https://user-images.githubusercontent.com/18056781/39081659-fdd4655e-4562-11e8-86f9-a5fab28634bf.JPG" />
+
+<img src="https://user-images.githubusercontent.com/18056781/39081665-10eda006-4563-11e8-9455-986b5a23934e.jpg" />
+
+- Press 'y' to mark the images as compatible.
```
marco-c: `;` at the end of every list item, except for `.` for the last item.

PR #180, opened 2018-05-11
Add text file for train.py

sagarvijaygupta:
Save machine information.
Save network information with Accuracy.


sagarvijaygupta:
@marco-c Below are the key points

1.  `file_name` is created in the following format `name_of_user+timestamp`.
2. Machine information is added using `lscpu`. Not all parameters are required to be in the file. I am not sure which all should I remove. 
3. Also I am thinking to keep the `filename` as `argparse` also if user wishes for the same.
4. I think we should change `argparse` interface in following format 

> python train.py --network=vgg16 --optimizer=adam --classification_type='Y vs D + N'

or shorter 

>  python train.py -n=vgg16 -o=adam -ct='Y vs D + N'

rather having 

>  python train.py vgg16 adam 'Y vs D + N'

marco-c:
> I think we should change argparse interface in following format

I agree, can you open an issue for this? We should support both the long format ("--network=") and the short one ("-n="), this should be feasible with argparse.

marco-c:
The main information is actually the GPU, the CPU will not matter that much.

marco-c:
We should have a link between the info file, the created model and the score.

sagarvijaygupta:
> We should have a link between the info file, the created model and the score.

I didn't get you. What type of link?

marco-c:
> I didn't get you. What type of link?

I mean, it should be possible to know to which saved model the info belongs to. E.g. by giving them the same name but with a different extension, or something like that.

autowebcompat/utils.py
```diff
@@ -226,3 +228,24 @@ def read_bounding_boxes(file_name):
...
+    for row in lscpu:
+        row = row.split(':')
+        parameter_value_map[row[0]] = row[1].strip()
+    return parameter_value_map
+
+
+def write_train_info(information, file_name=None):
+    if file_name is None:
+        file_name = subprocess.check_output('uname -n', shell=True).strip().decode()
+        file_name += datetime.now().strftime('_%H_%M_%Y_%m_%d.txt')
```
marco-c: The machine info is unlikely to change for the same user, so perhaps we should just use the user name.

sagarvijaygupta: For the filename? If a user tries different models one after the other then filenames should be different.

marco-c: If we store other information and not only the machine info, yes.

marco-c: I didn't realize this was still a WIP, sorry.

sagarvijaygupta: @marco-c Sorry I forgot to mark it as WIP.
Presently I am storing 
network, optimizer, classification_type, early_stopping, Accuracy, GPU1, GPU2, CPU(s), Thread(s) per core, Core(s) per socket, Model name.
Should I add anything or remove something?

marco-c: Sounds good, we can add more if we thing of something.

sagarvijaygupta: We can add time taken to train with number of training samples and testing time with number of samples also.

sagarvijaygupta: @marco-c I was thinking to add model summary to the file. Is it a good idea? We are already saving the name of the model though!

sagarvijaygupta: @marco-c I have added the model summary part. Since we were returning a model from the function, I have implemented a recursive approach so that even if we join multiple models in future we can get their summary.

train.py
```diff
@@ -11,10 +11,10 @@
 random.seed(42)
 
 parser = argparse.ArgumentParser()
-parser.add_argument('network', type=str, choices=network.SUPPORTED_NETWORKS, help='Select the network to use for training')
```
marco-c: We should do this in a separate PR. Each PR should be self-contained.

sagarvijaygupta: Oh! Okay! I will remove it here.

sagarvijaygupta:
There is one thing I wanted to mention that the present code to write in file uses some functions which are linux specific! And won't work on mac or windows. Should I provide a check for it? I am not sure even if I try to implement it for mac and windows I will be able to test it or not!

marco-c:
> @marco-c I have made the necessary additions. Any more things you think I should add or change?

The information you're storing seems enough. Let me know when you think the PR is ready by removing the "[WIP]" tag from the title.

> There is one thing I wanted to mention that the present code to write in file uses some functions which are linux specific! And won't work on mac or windows. Should I provide a check for it? I am not sure even if I try to implement it for mac and windows I will be able to test it or not!

Just add a simple check to only add this info when Linux is being used.

train.py
```diff
@@ -59,12 +60,20 @@ def gen_func(images):
...
 network.compile(model, args.optimizer)
-
 callbacks_list = [ModelCheckpoint('best_train_model.hdf5', monitor='val_accuracy', verbose=1, save_best_only=True, mode='max')]
 
 if args.early_stopping:
     callbacks_list.append(EarlyStopping(monitor='val_accuracy', patience=2))
 
+start_train_time = time.time()
 model.fit_generator(train_iterator, callbacks=callbacks_list, validation_data=validation_iterator, steps_per_epoch=train_couples_len / BATCH_SIZE, validation_steps=validation_couples_len / BATCH_SIZE, epochs=EPOCHS)
+train_time = time.time() - start_train_time
```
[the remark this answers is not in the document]
sagarvijaygupta: @marco-c I did find a better way to implement time functionality by adding it in callbacks. Also with every epoch we will be able to store loss and accuracy which will surely be helpful in analyzing data.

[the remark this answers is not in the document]
marco-c: Good, let me know when you're done with the modifications!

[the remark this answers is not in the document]
sagarvijaygupta: @marco-c I have made the necessary changes.

marco-c:
There's a flake8 issue that should be fixed before I can merge this.

sagarvijaygupta:
@marco-c flake8 issue fixed.

PR #183, opened 2018-05-20
Simplify train.py and pretrain.py arguments

### LINKED ISSUE #182: Change argparse interface for train.py
So that `train.py` runs like:

>  python train.py --network=vgg16 --optimizer=adam --classification_type='Y vs D + N'
 python train.py -n=vgg16 -o=adam -ct='Y vs D + N'

Shashi456:
Fixes #182, adding the same for pretrain.py as well.

### DIFF
--- a/pretrain.py
+++ b/pretrain.py
@@ -14,9 +14,9 @@ EPOCHS = 50
 random.seed(42)
 
 parser = argparse.ArgumentParser()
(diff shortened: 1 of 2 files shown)


marco-c:
Can you update train.py too?

Shashi456:
Isn't train.py already updated ?

marco-c:
> Isn't train.py already updated ?

No.

PR #185, opened 2018-05-23
Convert pretrain and train scripts to notebooks

### LINKED ISSUE #181: Write notebooks that can be used for training
Either rewrite train.py and pretrain.py as notebooks, or create notebooks that import those scripts and run them.

Shashi456:
Fixes #181. 
I have simply converted the scripts to notebooks.


Shashi456:
@marco-c i definitely agree. Notebooks are very hard to review. But our original motive still stands strong , it's very true that notebooks are easier to run online. One major plus point is the cell, which makes debugging easier. In a novel project like ours, debugging is important as well.  Would you rather prefer to postpone this to a later point in time?  I would suggest instead of converting the pretrain script. We will have a notebook for just train.py for convenience. While the scripts would still exist for review purposes.

Or we could wait till we had a more concrete version of train.py.

marco-c:
What do you think of the other option I proposed? Creating notebooks that import the scripts and run them

Shashi456:
Do you mean importing the train.py script ? Wouldn't that need us to refactor train.py in a modular way ?

marco-c:
> Do you mean importing the train.py script ? Wouldn't that need us to refactor train.py in a modular way ?

Yes, the refactoring would be done as part of the move to notebooks.

marco-c:
> @marco-c can you look at the notebooks I've posted and tell me if the cells are rightly demarcated, if yes I'll convert the required cells to functions in the original script.

They're more or less OK. The arg parse stuff should be removed (as it doesn't make sense in a notebook), and I would split the network creation from its training and its evaluation.

marco-c:
> @marco-c when i remove the arg parse , we will no longer be allowed to use the train.py script with the dynamism we wish to use it with. We will just have to manually change each time we wish for a change, train.py will then be something we just use as a file for the notebooks.

I meant that the arg parse should be removed from the notebook.

Shashi456:
But we instead of copying the code we will import the train script right? if we keep the argparse in the train script and try to import that into the notebook we get a ```ipykernel_launcher``` error

marco-c:
> But instead of copying the code we will import the train script right? if we keep the argparse in the train script and try to import that into the notebook we get a ipykernel_launcher error/.

We can move the train.py code into another script. Then have:
- A common script with the functionality in functions;
- A notebook which loads the common script;
- A script that runs in the command line (with arg parse) which loads the common script.

> And should i close this pr ? since #186 solved our issue currently?

#186 added a basic notebook, which makes this issue less important, but still it would be nicer to have a cleaner notebook split in multiple sections with explanations.
So, the priority is low, if you want to work on something else you're welcome to :)

marco-c:
Let's close this, since we have basic notebooks that are working.

PR #186, opened 2018-05-29
Load repository in Google Colab

sagarvijaygupta:
To run different models on Google Colab using GPU

### DIFF
--- a/load_repo.ipynb
+++ b/load_repo.ipynb
@@ -1,0 +1,248 @@
+{
+  "nbformat": 4,
+  "nbformat_minor": 0,
+  "metadata": {
+    "colab": {
+      "name": "load_repo.ipynb",
+      "version": "0.3.2",
+      "views": {},
+      "default_view": {},
+      "provenance": []
+    }
+  },
+  "cells": [
+    {
+      "metadata": {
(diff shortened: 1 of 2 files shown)


load_repo.ipynb
```diff
@@ -0,0 +1,248 @@
...
+        "colab": {
+          "autoexec": {
+            "startup": false,
+            "wait_interval": 0
+          }
+        }
+      },
+      "cell_type": "code",
+      "source": [
+        "! wget https://github.com/git-lfs/git-lfs/releases/download/v2.4.2/git-lfs-linux-amd64-2.4.2.tar.gz"
```
marco-c: Nit: no space between `!` and the command here and in the following cases.

train.ipynb
```diff
@@ -0,0 +1,248 @@
...
+        "colab": {
+          "autoexec": {
+            "startup": false,
+            "wait_interval": 0
+          }
+        }
+      },
+      "cell_type": "code",
+      "source": [
+        "!pip install -r test-requirements.txt"
```
marco-c: We shouldn't need to install test-requirements.txt, but we can remove this in a follow-up. I'll file an issue.

PR #190, opened 2018-05-31
 force_single_line configuration for isort

### LINKED ISSUE #189: Use force_single_line configuration for isort
force_single_line makes diffs easier to read.

As part of this, we should probably also remove `line_length=150`.

Shashi456:
Fixes #189.

### DIFF
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -7,12 +7,12 @@ import subprocess
 import sys
 import threading
 
-import numpy as np
+from PIL import Image
 import keras
 from keras.preprocessing.image import ImageDataGenerator
 from keras.preprocessing.image import img_to_array
 from keras.preprocessing.image import load_img
-from PIL import Image
+import numpy as np
 
 
 def get_bugs():
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -7,10 +7,12 @@ import subprocess
 import sys
 import threading
 
-from PIL import Image
-import keras
-from keras.preprocessing.image import ImageDataGenerator, img_to_array, load_img
 import numpy as np
+import keras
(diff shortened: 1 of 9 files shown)


marco-c:
There are still some flake8 failures (see Travis logs)

autowebcompat/utils.py
```diff
@@ -8,10 +8,13 @@
 import threading
 
 from PIL import Image
-import keras
-from keras.preprocessing.image import ImageDataGenerator, img_to_array, load_img
 import numpy as np
 
+import keras
```
marco-c: These imports should be in the previous section, as it's a third-party library

collect.py
```diff
@@ -8,10 +8,12 @@
 import traceback
 
 from PIL import Image
-from selenium import webdriver
-from selenium.common.exceptions import NoAlertPresentException, NoSuchWindowException, TimeoutException
 
 from autowebcompat import utils
+from selenium import webdriver
```
marco-c: Here too

label.py
```diff
@@ -1,9 +1,9 @@
 import argparse
 
-import cv2
 import numpy as np
 
 from autowebcompat import utils
+import cv2
```
marco-c: Here too

pretrain.py
```diff
@@ -4,9 +4,10 @@
 import random
 from urllib.parse import urlparse
 
-from keras.callbacks import EarlyStopping, ModelCheckpoint
-
-from autowebcompat import network, utils
+from autowebcompat import network
+from autowebcompat import utils
+from keras.callbacks import EarlyStopping
```
marco-c: Here too

Shashi456:
@marco-c  i will revise the errors, i directly executed isort on my terminal and it seems to have rearranged the files in an inappropriate way

PR #192, opened 2018-06-04
Change pip installations

sagarvijaygupta:
I didn't notice that after installing pip requirements training was using CPU instead of GPU.


train.ipynb
```diff
@@ -1,247 +1,249 @@
...
+          "autoexec": {
+            "startup": false,
+            "wait_interval": 0
+          }
+        }
+      },
+      "cell_type": "code",
+      "source": [
+        "%%writefile requirements.txt\n",
+        "selenium==3.4.3\n"
```
marco-c: So Keras and Tensorflow are already available?
Since `selenium` is the only needed package, you could just do `!pip install selenium==3.4.3`.

sagarvijaygupta: Yeah they are already available. Pre-installed for GPU version. I was just overwriting them with CPU version :disappointed: .  We need to install selenium only.

PR #193, opened 2018-06-08
Getting information of GPU using tensorflow.

sagarvijaygupta:
This will make sure that in the end we know whether GPU was used for training purpose or not.
Moreover `lshw` which was previously used might not always be available in `linux`.
Also we get GPU's `memory info`.


autowebcompat/utils.py
```diff
@@ -251,11 +252,14 @@ def get_machine_info():
     if 'linux' not in operating_sys:
         return parameter_value_map
 
-    gpu = subprocess.check_output('lshw -C display | grep product', shell=True).strip().decode()
-    gpu = gpu.split('\n')
+    gpu = []
+    for device in device_lib.list_local_devices():
+        if device.device_type == 'GPU':
+            gpu.append(device)
     for i in range(len(gpu)):
```
marco-c: This for loop is no longer needed, you can do:
```Python
for i, device in enumerate(device_lib.list_local_devices()):
    if device.device_type != 'GPU':
        continue

    parameter_value_map['GPU_{}_name'.format(i + 1)] = device.name
    ...
```

sagarvijaygupta: Done!

PR #196, opened 2018-06-10
Bug fix in label.py

sagarvijaygupta:
Bug fix where instead of pressing 'y' user presses `<Return>` twice.


marco-c:
Can you split the two things in separate PRs? One to add the shuffling, and the other to fix the bug.

sagarvijaygupta:
I modified this PR only for bug-fix.

PR #197, opened 2018-06-10
Images labelled with bounding boxes

sagarvijaygupta:
1190 images till now. Will continue to annotate them side by side while training continues.


marco-c:
Thanks! Could you remove the first commit from this PR? It should just be the second commit.
You can do this:
```
git checkout labels_sagar
git rebase -i HEAD~2
```
In the editor, add "d" for the first commit, then save the file.
Then: `git push origin labels_sagar --force` to update the PR (you need the `--force` because you have modified the history of the branch`).

PR #201, opened 2018-06-11
Create train_info folder if it doesn't exist.

sagarvijaygupta:


autowebcompat/utils.py
```diff
@@ -272,6 +272,9 @@ def write_train_info(information, model, train_history, file_name=None):
         file_name += datetime.now().strftime('_%H_%M_%Y_%m_%d.txt')
     machine_info = get_machine_info()
     information.update(machine_info)
+
+    if not os.path.exists('train_info'):
```
marco-c: Use the `exist_ok` parameter of `os.makedirs` instead of checking if the directory exists first.

PR #202, opened 2018-06-12
Convert train.py script directly into notebook.

sagarvijaygupta:
Update notebook so that model remains saved and can be used further for prediction and creating confusion matrix.
Instead of `argparse` interface we have a `dictionary` named `args` which takes values similarly as `argparse` would take.

### DIFF
--- a/train.ipynb
+++ b/train.ipynb
@@ -1,230 +1,399 @@
 {
-  "nbformat": 4,
-  "nbformat_minor": 0,
-  "metadata": {
-    "colab": {
-      "name": "train.ipynb",
-      "version": "0.3.2",
-      "views": {},
-      "default_view": {},
-      "provenance": []
-    }
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {
+    "colab_type": "text",
+    "id": "JQwDy3kmIur6"
+   },
+   "source": [
+    "*Downloading git LFS*"
+   ]
   },
-  "cells": [
-    {
-      "metadata": {
-        "id": "JQwDy3kmIur6",
-        "colab_type": "text"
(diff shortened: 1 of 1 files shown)


Shashi456:
@sagarvijaygupta  so i tried to do this in #185, we decided that the codes would need to be refactored in order to be written as notebooks. Since notebooks are harder to review and we would be having some substantial changes in train.py , the refactored code would be more modular i.e functions which can be called in the notebook since that makes it convenient to read. As of now converting train.py to a notebook is low priority because of #200 and #186

sagarvijaygupta:
@Shashi456 Okay! I made it for the people who are working on Google Colab, so that they can use this to work on. Even after training finishes in this case we won't lose our model variables which is the main thing where notebooks help!

marco-c:
Can't we do the needed changes in train.py directly?

sagarvijaygupta:
@marco-c Yeah we can directly make the changes in train.py, but sometimes it is like we just want the model. After the training is finished to try and implement confusion matrix or other accuracy measures we can simply do it in new cells. If we make the changes in train.py we have to wait for the whole training to complete once again.

marco-c:
> @marco-c Yeah we can directly make the changes in train.py, but sometimes it is like we just want the model. After the training is finished to try and implement confusion matrix or other accuracy measures we can simply do it in new cells. If we make the changes in train.py we have to wait for the whole training to complete once again.

If you want to just test something, you can always modify the train.py script, no?

sagarvijaygupta:
@marco-c :smile: I get it! Yeah reviewing seems actually impossible using diffs. I will close the PR in that case. Thanks!

PR #203, opened 2018-06-15
Convert VGG16 model to pre-defined keras model.

sagarvijaygupta:


autowebcompat/network.py
```diff
@@ -45,42 +46,7 @@ def create_mlp(input_shape):
...
-    x = Conv2D(512, (3, 3), activation='relu', padding='same')(x)
-    x = MaxPooling2D(pool_size=(2, 2), strides=(2, 2))(x)
-
-    x = Flatten()(x)
-    x = Dense(4096, activation='relu')(x)
-    x = Dropout(0.5)(x)
-    x = Dense(4096, activation='relu')(x)
-    # Softmax layer Not Necessary
-    return Model(input, x)
+    return VGG16(input_shape=input_shape)
```
marco-c: One problem here is that we don't want the softmax layer

sagarvijaygupta: Oh! I will try to remove it.

autowebcompat/network.py
```diff
@@ -45,42 +46,9 @@ def create_mlp(input_shape):
...
-    x = MaxPooling2D(pool_size=(2, 2), strides=(2, 2))(x)
-
-    x = Flatten()(x)
-    x = Dense(4096, activation='relu')(x)
-    x = Dropout(0.5)(x)
-    x = Dense(4096, activation='relu')(x)
-    # Softmax layer Not Necessary
-    return Model(input, x)
+    base_model = VGG16(input_shape=input_shape)
+    base_model.layers.pop()
```
marco-c: It looks a bit of a hack, but I guess it's fine.

sagarvijaygupta: We can do 
> return Model(input=base_model.input, output=base_model.get_layer('fc2').output)

or 
> return Model(input=base_model.input, output=base_model.get_layer('predictions').input)

if that looks cleaner :smile:

marco-c: It's more or less the same, let's use the current code you wrote.

sagarvijaygupta: @marco-c I was trying to run the above model but it was giving some error while trying to save the model. I had tested it with the model summary and until first epoch starts. But since model was saved in the end of first epoch I never noticed it (it was unexpected).

Solution is the way I proposed in above comment ie, instead of using
> base_model.layers.pop() 

we use
>Model(input=base_model.input, output=base_model.get_layer('fc2').output)

PR #206, opened 2018-06-15
Add pretrained weights on imagenet for vgg16, vgg19, resnet50

sagarvijaygupta:


sagarvijaygupta:
Resolved!

train.py
```diff
@@ -16,6 +16,7 @@
 parser = argparse.ArgumentParser()
 parser.add_argument('-n', '--network', type=str, choices=network.SUPPORTED_NETWORKS, help='Select the network to use for training')
 parser.add_argument('-o', '--optimizer', type=str, choices=network.SUPPORTED_OPTIMIZERS, help='Select the optimizer to use for training')
+parser.add_argument('-w', '--weights', type=str, choices=network.SUPPORTED_WEIGHTS, help='Select the weights to be loaded for given model')
```
marco-c: `for the given model`

autowebcompat/network.py
```diff
@@ -185,22 +187,26 @@ def create_inception_network(input_shape):
...
-    base_network = network_func(input_shape)
+    base_network = network_func(input_shape, weights)
 
     input_a = Input(shape=input_shape)
     input_b = Input(shape=input_shape)
 
     # Loading pretrained weights corresponding to the network used
-    if weights:
-        base_network.load_weights(weights)
+    # if weights:
```
marco-c: Why commented out?

sagarvijaygupta: Because if we load weights over here we have to specify the location of weights. For imagenet keras automatically downloads them.

marco-c: We should support both downloading by Keras and allowing custom weights to be loaded

sagarvijaygupta: Done!

autowebcompat/network.py
```diff
@@ -185,23 +187,29 @@ def create_inception_network(input_shape):
...
+def create_resnet50_network(input_shape, weights):
+    base_model = ResNet50(input_shape=input_shape, weights=weights)
     return Model(inputs=base_model.input, outputs=base_model.get_layer('flatten_1').output)
 
 
 def create(input_shape, network='vgglike', weights=None):
     assert network in SUPPORTED_NETWORKS, '%s is an invalid network' % network
+    if weights is not None:
+        assert network in SUPPORTED_NETWORKS_WITH_WEIGHTS, '%s does not have weights for %s ' % (network, weights)
+        assert input_shape == (224, 224, 3), 'shape must be (224, 224, 3)'
```
marco-c: This should be removed as it is actually unrelated to this PR

autowebcompat/network.py
```diff
@@ -185,23 +187,29 @@ def create_inception_network(input_shape):
...
-    base_model = ResNet50(input_shape=input_shape)
+def create_resnet50_network(input_shape, weights):
+    base_model = ResNet50(input_shape=input_shape, weights=weights)
     return Model(inputs=base_model.input, outputs=base_model.get_layer('flatten_1').output)
 
 
 def create(input_shape, network='vgglike', weights=None):
     assert network in SUPPORTED_NETWORKS, '%s is an invalid network' % network
+    if weights is not None:
+        assert network in SUPPORTED_NETWORKS_WITH_WEIGHTS, '%s does not have weights for %s ' % (network, weights)
```
marco-c: When `weights` is a file, we shouldn't check if the network is in SUPPORTED_NETWORKS_WITH_WEIGHTS, as it could be any of the networks

autowebcompat/network.py
```diff
@@ -185,23 +187,29 @@ def create_inception_network(input_shape):
...
+        assert input_shape == (224, 224, 3), 'shape must be (224, 224, 3)'
+        weights_file = None
+        if weights != 'imagenet':
```
marco-c: weights not in SUPPORTED_WEIGHTS

marco-c:
I think the best option to simplify the implementation is to have two parameters, `builtin_weights` and `weights`.

The first parameter allows to choose between SUPPORTED_WEIGHTS.
The second parameter is a path to a file.

Only one of the two can be not None (you can add an assertion at the beginning of the function).

autowebcompat/network.py
```diff
@@ -185,15 +187,20 @@ def create_inception_network(input_shape):
...
 
 
-def create(input_shape, network='vgglike', weights=None):
+def create(input_shape, network='vgglike', weights=None, builtin_weights=None):
     assert network in SUPPORTED_NETWORKS, '%s is an invalid network' % network
+    assert weights is None or builtin_weights is None, 'only one type of weights are allowed at once'
+
+    if builtin_weights:
+        assert network in SUPPORTED_NETWORKS_WITH_WEIGHTS, '%s does not have weights for %s ' % (network, builtin_weights)
+
```
marco-c: You could also assert that builtin_weights is in SUPPORTED_WEIGHTS.

sagarvijaygupta: I think that is already asserted when I pass it as argparse using `choice=network.SUPPORTED_WEIGHTS`

marco-c: That's just the user interface. In the future there could be other things calling the library without going through the user interface, so it's good to program "defensively".

PR #207, opened 2018-06-15
[WIP] Testing VGG19, VGG16, and VGG-like (from scratch versions)

sdv4:
Will include testing various parameters for the architecture, and labelling more images.


sagarvijaygupta:
You must change VGG19 model to

```
def create_vgg19_network(input_shape):
    base_model = VGG19(input_shape=input_shape)
    return Model(input=base_model.input, output=base_model.get_layer('fc2').output)
```

There was some unseen error in saving of models. I will create a PR for this shortly.

sdv4:
@sagarvijaygupta I made that change now, thanks!

marco-c:
@sdv4 are you labeling with bounding boxes too? It is optional, but it would be better as it would give us more information about the incompatibilities.

sdv4:
@marco-c Yes, I will do that. 
Just fyi: the first run resulted in validation accuracy of 91.19% after the first epoch, with no improvement by the fourth epoch when I killed the process (was just running locally on my machine).

marco-c:
@sdv4 could you split the labeling into a separate PR, so we can merge it?

marco-c:
And we should look at the confusion matrix, that percentage looks too good to be true!

sdv4:
@marco-c Will do!

As for the accuracy percentages, were you expecting them to be much lower because the networks aren't using the ImageNet weights?
How should I go about investigating this further?

marco-c:
> As for the accuracy percentages, were you expecting them to be much lower because the networks aren't using the ImageNet weights?

I was pessimistic, the problem is not so easy so I was expecting some tuning to be needed.

> How should I go about investigating this further?

@sagarvijaygupta added support for outputting the confusion matrix, the first step would be to retrain with one of the networks and see what the confusion matrix looks like.

sdv4:
@marco-c 
I did include the output confusion matrices:

VGG19:
[[136 39]
[ 26 215]]

VGG16:
[[142 61]
[ 22 191]]

Is this what you were referring to?

marco-c:
> Is this what you were referring to?

Yes, sorry, I had missed them, that issue was getting too long :P
They don't look too imbalanced!

PR #209, opened 2018-06-16
Add confusion matrix for test iterator.

sagarvijaygupta:
I tried directly using the `predict_generator` available in `keras` to get the predicted values. I was not able to get the `true values` from it. I tried adding an accumulator in `CouplesIterator` class and append `y_batch` to it, but `next` function was being called 23 times which should have been 12 times (using `steps=test_couples_len / BATCH_SIZE`) `test_couples_len = 396`

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,4 +3,5 @@ selenium==3.4.3
 keras==2.1.1
 tensorflow==1.3.0
 h5py==2.7.1
-opencv-python==3.4.0.12 
+opencv-python==3.4.0.12
+scikit-learn==0.19.1 
--- a/train.py
+++ b/train.py
@@ -5,6 +5,8 @@ import time
 from keras.callbacks import Callback
 from keras.callbacks import EarlyStopping
 from keras.callbacks import ModelCheckpoint
+import numpy as np
+from sklearn.metrics import confusion_matrix
 
 from autowebcompat import network
 from autowebcompat import utils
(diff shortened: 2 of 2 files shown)


marco-c:
> I tried directly using the predict_generator available in keras to get the predicted values. I was not able to get the true values from it.

What was the problem exactly?
There's this example from which you could get inspiration: https://gist.github.com/RyanAkilos/3808c17f79e77c4117de35aa68447045.

sagarvijaygupta:
@marco-c I did see that example. The problem was that I needed `true` values of the labels for confusion matrix. Since we created our own custom iterator I was not sure how to get those `true` values.
 
What I did was, I took a variable `y_true` inside the `CoulesIterator` class, and extend the `y_batch` to it (`self.y_true.extend(y_batch`) (initialising it to empty in `__iter__`). 

When I tried to create confusion matrix from the `y_pred` (`y_pred = model.predict_generator(test_iterator, steps=test_couples_len / BATCH_SIZE)`) and `y_true` (`y_true = test_iterator.y_true`) I got size mismatch. 

`y_true` had a size of `736` while `y_pred` had `416`. I did understand why `416` as we had `396` couples so it took them in batches of `32` and in the last batch it took some repetitions. But I didn't understand how the size of `y_true` came to be `736`.

marco-c:
Doesn't model.predict_generator called with our custom iterator work? In theory they should have the same interface, so they should be interchangeable. All the other functions (e.g. `evaluate_generator`) work, so I would expect this one to work too.

marco-c:
> Doesn't model.predict_generator called with our custom iterator work? In theory they should have the same interface, so they should be interchangeable. All the other functions (e.g. evaluate_generator) work, so I would expect this one to work too.

Forget this, I see it worked. Your problem was how to get y_true.

sagarvijaygupta:
@marco-c if you find a reason for the problem, please share it with me. It actually bugged me for quite some time. :smile:

PR #216, opened 2018-06-19
Add functionality to change previously labeled images.

sagarvijaygupta:
Press 'a' to go to previous image.
Press 'd' to go to next image.

### DIFF
--- a/label.py
+++ b/label.py
@@ -17,12 +17,15 @@ bounding_boxes = utils.read_bounding_boxes(labels_directory + args.file_name + '
 
 images_to_show = [i for i in utils.get_images() if i not in labels]
 random.shuffle(images_to_show)
+images_marked = [i for i in utils.get_images() if i in labels]
+images_to_show = images_marked + images_to_show
+image_index = len(images_marked)
 drawing = False
 shifting = False
 changing_shape = False
 box_to_change = {}
 all_boxes = {}
-key_map = {'Escape': 27, 'r': 114, 'Enter': 13, 'Space': 32, 'y': 121, 'n': 110, 'd': 100}
+key_map = {'Escape': 27, 'r': 114, 'Enter': 13, 'Space': 32, 'y': 121, 'left_a': 97, 'right_d': 100}
 cv2.namedWindow('firefox')
 cv2.namedWindow('chrome')
 cv2.namedWindow('firefox_chrome_overlay')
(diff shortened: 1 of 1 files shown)


label.py
```diff
@@ -116,6 +119,27 @@ def create_change_shape(drawing_area, start_x, start_y, end_x, end_y):
     cv2.arrowedLine(drawing_area, (end_x - 2, end_y - 2), (end_x - 12, end_y - 12), (255, 0, 0), 2, tipLength=0.3)
 
 
+def draw_bounding_boxes_init(param):
+    [main_drawing_area, secondary_drawing_area, boxes] = param
+    # GREEN --> 'n'     YELLOW --> 'd'
+    color = {'d': (0, 255, 255), 'n': (0, 255, 0)}
```
marco-c: Define the colors at the beginning as constants, so you can change them for all places in one go.
COLOR_N = (0, 255, 0)  # GREEN
COLOR_D = (0, 255, 255)  # YELLOW

label.py
```diff
@@ -17,12 +17,15 @@
...
+images_marked = [i for i in utils.get_images() if i in labels]
+images_to_show = images_marked + images_to_show
+image_index = len(images_marked)
 drawing = False
 shifting = False
 changing_shape = False
 box_to_change = {}
 all_boxes = {}
-key_map = {'Escape': 27, 'r': 114, 'Enter': 13, 'Space': 32, 'y': 121, 'n': 110, 'd': 100}
+key_map = {'Escape': 27, 'r': 114, 'Enter': 13, 'Space': 32, 'y': 121, 'left_a': 97, 'right_d': 100}
```
marco-c: Add the new commands to the help.

sagarvijaygupta: Should I do it in the same PR?

label.py
```diff
@@ -17,12 +17,15 @@
 
 images_to_show = [i for i in utils.get_images() if i not in labels]
 random.shuffle(images_to_show)
+images_marked = [i for i in utils.get_images() if i in labels]
+images_to_show = images_marked + images_to_show
```
marco-c: We should add an option to either only show unmarked images or show all images.

label.py
marco-c: Yes, since you're adding the commands here, it makes sense to also update the help here.

label.py
sagarvijaygupta: Presently it starts with the very next image you would have started labeling. If you press 'a' (to go left) it will show all the previously marked images. 
I can add it as an argparse interface if you feel it is necessary!

marco-c: Yeah it's a bit confusing, I think it's better to add two modes

sagarvijaygupta: Okay! I will add two modes. One only for unmarked images, and other for all images including unmarked images?

marco-c: Maybe one for unmarked and one for marked? It sounds even better (in one case, you only want to verify something you or somebody else has marked, in the other case you want to mark new images).

sagarvijaygupta: Just to confirm, user will be able to edit his old labels. 
So in case of marked, user will be able to see his previous markings and edit them. He won't be allowed to mark new images.
In case of unmarked, user will be able to mark new images and change only the images he has marked in that run.

sagarvijaygupta:
@marco-c If this feels right, i will update the help part.

label.py
marco-c: Exactly!

marco-c:
> @marco-c If this feels right, i will update the help part.

Yes, looks good!

PR #220, opened 2018-06-21
Add readme about dataset labeling

sagarvijaygupta:


labeling_guide/ReadMe.md
```diff
@@ -0,0 +1,56 @@
+
+# Labeling Guidelines
+
+## Images marked as compatible - y 
+
+---
+
+#### They look same.
```
marco-c: `They look the same`

labeling_guide/ReadMe.md
```diff
@@ -0,0 +1,56 @@
+
```
marco-c: We should move everything about labeling in this document, put it in the top-level directory calling it "LABELING.MD", and referencing it in README.md.

labeling_guide/ReadMe.md
```diff
@@ -0,0 +1,56 @@
...
+# Labeling Guidelines
+
+## Images marked as compatible - y 
+
```
marco-c: Let's remove the newlines around the --- lines.

marco-c: (if possible)

labeling_guide/ReadMe.md
```diff
@@ -0,0 +1,56 @@
...
+----
+
+
+## Bounding boxes marked as different yet compatible - d 
+
+---
+
+#### They look different.
+#### Mark the parts which are same logically with Yellow.
+>Different advertisements, different videos loaded, time-in-clock are marked
```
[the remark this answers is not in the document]
sagarvijaygupta: A screenshot which has present time displayed, it will be different in two screenshots as they are taken at different times but still compatible.

LABELING.md
```diff
@@ -0,0 +1,51 @@
...
+> Improper loading of images, missing text, different design, different languages are marked incompatible.
+---
+
+## Bounding boxes marked as different yet compatible - d 
+---
+#### Couples of images that are compatible, but with content differences.
+#### They look different.
+#### Mark the parts which are logically the same.
+>Different advertisements, different videos loaded, time-in-clock, captchas are marked
+as different yet compatible.
```
marco-c: Also add the example which was in the readme about the news site.

`Different videos loaded` shouldn't be there, the same video should be loaded in both browsers if the site is compatible. You can say `different frames of a video loaded`.
Rename `time-in-clock` as it is unclear.

I would say something like `Since the screenshots are taken at different times in the two browsers, there are differences which are not incompatibilities but are actually due to the different timing. For example, a banner could be showing a different advertisement, a video could be in two different frames, a clock could be showing different time, a captcha could be showing different characters or images, two news could be different, and so on`.

LABELING.md
```diff
@@ -0,0 +1,51 @@
...
+<p align="center"><img src="labeling_guide/y3.png" width=617 height=357></p>
+<p align="center"><img src="labeling_guide/y4.png" width=617 height=357></p>
+<p align="center"><img src="labeling_guide/y5.png" width=617 height=357></p>
+
+## Bounding boxes marked as incompatible - n 
+---
+#### Couples of images which are not compatible
+#### They are different.
+#### Mark the parts which are logically different.
+> Improper loading of images, missing text, different design, different languages are marked incompatible.
```
marco-c: For example, improper loading ...

Add `for example` because these are not the only cases.

marco-c:
Thanks! I missed you made the updates to the PR... There are some conflicts though.

marco-c:
The colors are wrong now, as we changed them in #221.

sagarvijaygupta:
@marco-c I will have to update the images in that case!

sagarvijaygupta:
@marco-c I have changed the images accordingly.

PR #226, opened 2018-06-27
Add dom based compatibility with a basic technique.

sagarvijaygupta:
Brief technique to walk the DOM in both Firefox and Chrome and see if there are:
1) elements in Firefox that are not in Chrome (and the other way around);
2) for elements that are in both browsers, check if they have different properties.

Just a basic prototype:
I ran it for 191 doms, for which it marked only 15 as compatible according to our above technique.


marco-c:
Let's analyze a couple of the examples marked as incompatible to see where we can improve.
Have you calculated the accuracy? (how many of those incompatible are actually incompatible)

sagarvijaygupta:
@marco-c I analyzed many images. Screenshots marked compatible were compatible only as expected. But most of the incompatibilities are false. The screenshots are same visually.

marco-c:
Let's take a few incompatibilities (say, 7) and check why they are marked as incompatible by this technique (e.g. maybe there's something trivial that is causing many of them to be marked as incompatible).

marco-c:
By "why" I mean: is there an element which is in one but not in the other? If yes, which element? If no, is there a property of an element which is different? Which property? What are the two different values?

sagarvijaygupta:
I had put print statements in the comparator function. Many of them showed number of nodes mismatched. Some had number of attributes mismatched. I will do a thorough analysis, and then post the results.

marco-c:
> I had put print statements in the comparator function. Many of them showed number of nodes mismatched. Some had number of attributes mismatched. I will do a thorough analysis, and then post the results.

OK, thanks! Let's start with a single example of compatible wrongly marked as incompatible.

sagarvijaygupta:
@marco-c I am thinking of trying `tree edit distance` as a similarity parameter for comparing dom trees.

marco-c:
> @marco-c I am thinking of trying tree edit distance as a similarity parameter for comparing dom trees.

Then you'd need to have a threshold, I'm not sure it will work well. I'd look more at the paper (and associated source code) I sent you to understand what they were doing. E.g. maybe they are considering only some attributes.

Let's analyze a few cases first. Why are they considered incompatible? For 5764, what's the missing xmlns:xlink in Chrome? The src values are not exactly the same, but we definitely need to ignore them (startTime, endTime, guid, are definitely going to be different, but it doesn't matter).

sagarvijaygupta:
@marco-c I am also sceptical  about the tree idea.

On analyzing I am not even sure whether they are visible in the screenshot or not. There is a text `Learn more about Netflix below` which is not visible in the screenshot but is there is in DOM. Below are the differences in the both when seen in DOM file. These lines are just after the above text so we can't analyze them on the basis of screenshot. 
```
Chrome -    <use filter="" xmlns:xlink="http://www.w3.org/1999/xlink" xlink:href="#netflix-logo"></use>
Firefox -   <use filter="" xlink:href="#netflix-logo"></use>
```
For src values its clear that they are similar and we will have to ignore them while testing.

For the `iFrame` and `img` one, below `img` tag is not there in `chrome`.
```
<img src="/ichnaea/cl2/freeform/DebugEvent?source=www&amp;action=jsSimplicityPrefetch&amp;statusCode=200&amp;startTime=1529849885429&amp;endTime=1529849886164" style="display: none">
```
This last thing is clearly an incompatibility but this part is also not visible in screenshot. This part is loaded on screen but seems to be hidden until triggered by some button click.

marco-c:
We should probably ignore everything that is hidden (e.g. "display: none").

marco-c:
We should probably ignore everything that is hidden (e.g. "display: none").

sagarvijaygupta:
@marco-c That should be done in crawler itself?

marco-c:
> @marco-c That should be done in crawler itself?

We can do it either in the crawler or in the script to find differences.
Doing it in the crawler might be easier.

sagarvijaygupta:
How can we do it in the script? 
For the crawler part, we will get the same dom tree as we are getting now but we will store the displayed elements in a separate file?

marco-c:
> How can we do it in the script?

E.g. if there's 'display: none' then it's not a visible element. But it might take a while to get everything right!

> For the crawler part, we will get the same dom tree as we are getting now but we will store the displayed elements in a separate file?

We could store the dom tree in a file, and additional attributes we calculate (for now just "is visible", in the future other attributes) in a support file.

Have you looked at what https://github.com/gatech/xpert is doing? Maybe we can use the same properties they are using and ignore the same properties they are ignoring, instead of redoing the work completely from scratch.

sagarvijaygupta:
@marco-c I did go through the code once, they are ignoring `style and type` attributes  
1. `LevenshteinDistance` between xpaths
2. map similarity among attributes and doms.

sagarvijaygupta:
I am summarizing what I understood from the code:

For matching a node `ref` with every node `node` they have three criterias:

1. They are comparing `xpaths`using `LevenshteinDistance` ( `1 - getLevenshteinDistance(xPath1, xPath2)/max(xPath1.length(), xPath2.length())`). _(given weight 70%)_.

2. Comparing attributes of both (after removing `type` and `style`) by checking if both have the key and their values are matched as string by replacing  _`', ", whitespace`_  with `empty`. All matched attributes are added (even twice) and a value is returned `total_matched/sum_of_size_of_attribute_map`._(given weight 20%)_.

3. Comparing `Clickable` and `Visible` parameters of the node. _(given weight 10%)_

Using these criteria they have three levels of matching:

1. Perfect Matching 
Where the `Threshold` is set to `1` for `MatchIndex` (got from above).

2. Level Matching
Where the `Threshold` is set to `0.75` given the level of the node (level assigned using BFS) are same for both nodes.

3. Approximate global matching
Where the `Threshold` is set to `0.85` 

After this they return three sets of nodes `matched`, `unmatched1` and `unmatched2`.
Next is the `AlignmentGraph` which I will comment after understanding.

sagarvijaygupta:
In the `AlignmentGraph` they check for the matched nodes: 

1. They create an `AlignmentGraph` for two doms where they create a `vMap` (mapping from xPath to node), then sort the vertices based on Area and DOM hierarchy, and add nodes as a pair of `(parent, node)` when parent contains the node (node is inside the parent visually and considering large number of factors). Similarly add `siblings` who have same parent.

2. For all pair of matched nodes they compare their parents and siblings for large variety of criteria including `different alignments, different parents, vfill, hfill` etc considering many size differences greater than some threshold for both parents and siblings. These return large number of issues to the checker.

sagarvijaygupta:
Even after implementing them together we still need to think about something to increase the speed. 
Though we have less than 10 cases for 191 test cases where sum number of number of nodes is greater than 1700.
I am attaching the results file.
[out.txt](https://github.com/marco-c/autowebcompat/files/2151987/out.txt)

sagarvijaygupta:
@marco-c First matcher part is ready. I think we should analyse it first before moving to next part. They have not added `src` and `href` to ignored attribute. Should I add it? 
Also matching screenshots with results from this won't be right until we get the full screenshot. That is with whole page with scrolling.
```
Unmatched Nodes Chrome:
/html/body/script[4]
{'src': '//ethn.io/mob/41183.js?guid=Q3FYCI5Z75HMTIFTENP7UXSFLA'}
----------
/html/body/img[2]
{'src': '/ichnaea/cl2/freeform/DebugEvent?source=www&action=jsSimplicityPrefetch&statusCode=200&startTime=1529850175855&endTime=1529850177087'}
----------
/html/body/img[1]
{'src': '/ichnaea/cl2/freeform/DebugEvent?source=www&action=cssSimplicityPrefetch&statusCode=200&startTime=1529850175855&endTime=1529850176439'}
----------
***************************************
Unmatched Nodes Firfox:
/html/body/img[1]
{'src': '/ichnaea/cl2/freeform/DebugEvent?source=www&action=cssSimplicityPrefetch&statusCode=200&startTime=1529849885430&endTime=1529849885817'}
----------
/html/body/img[2]
{'src': '/ichnaea/cl2/freeform/DebugEvent?source=www&action=jsSimplicityPrefetch&statusCode=200&startTime=1529849885429&endTime=1529849886164'}
----------
/html/body/script[4]
{'src': '//ethn.io/mob/41183.js?guid=NHJ372D74NERRFQH3KWTVLSTFY'}
----------
Chrome Nodes : 445
Firefox Nodes : 445
Matched Nodes : 442
```

marco-c:
Thanks! Great investigation!
The speed maybe is not too much of a problem, we can parallelize it (each website comparison running in its own thread).
I agree we should look at some results first before moving forward.

Actually I was thinking we could even look at the results of their tool directly and see how often it detects incompatibilities in our dataset (we would need to re-crawl though with their crawler, as their crawler is storing different format of data) and how often it detects incompatibilities when there are no incompatibilities (my fear is that this is going to happen often).

marco-c:
> Actually I was thinking we could even look at the results of their tool directly and see how often it detects incompatibilities in our dataset (we would need to re-crawl though with their crawler, as their crawler is storing different format of data) and how often it detects incompatibilities when there are no incompatibilities (my fear is that this is going to happen often).

What do you think of this?

sagarvijaygupta:
@marco-c I will try to run their tool directly (it is made for Windows). 
There is one problem though we won't be able to crawl the same way I guess we have crawled (it will require some good amount of changes if possible).  I will research on it more.

> how often it detects incompatibilities when there are no incompatibilities (my fear is that this is going to happen often).
It is correct mildly as far as I have seen. In the first part only, it detects some incompatibilities (even if they are not present). In the next part (which I have not implemented yet we will get more incompatibilities only).

sagarvijaygupta:
@marco-c  I am attaching the results of directly running the their app using Firefox (version-30) Chrome (version-67) for two websites` (https://mvnrepository.com/, https://stackoverflow.com/)`
[output.zip](https://github.com/marco-c/autowebcompat/files/2172601/output.zip)

marco-c:
> @marco-c I am attaching the results of directly running the their app using Firefox (version-30) Chrome (version-67) for two websites(https://mvnrepository.com/, https://stackoverflow.com/)a

There were no incompatibilities for these websites?

marco-c:
> There were no incompatibilities for these websites?

I mean, were there actual incompatibilities? I see the outputted lists of incompatibilities are pretty long.

sagarvijaygupta:
Yeah there were incompatibilities in both website, but I am pretty sure the reason is the version difference (which corresponds to nearly 4 years). I will try to run it on latest firefox, then only results will be worth comparing.

sagarvijaygupta:
Result for google.com
[issues.zip](https://github.com/marco-c/autowebcompat/files/2185959/issues.zip)

sagarvijaygupta:
I have uploaded the complete output results in link below for three websites.

[https://drive.google.com/open?id=17UmmXaaVPmPy94WARINlgcXAZAEnYZxz](url)

sagarvijaygupta:
SInce xpert provides issues, can't we have something like this:

`nodes_matched/max(nodes in firefox and chrome)` as the similarity measure and match it with `y_pred` instead of `accuracy` in 
https://github.com/marco-c/autowebcompat/blob/8da7c8d15985585911f82e9ff386efe109bdc16d/autowebcompat/network.py#L234-L237

marco-c:
> SInce xpert provides issues, can't we have something like this:

> nodes_matched/max(nodes in firefox and chrome) as the similarity measure and match it with y_pred instead of accuracy in

We should do the same as X-Pert. When does it consider two websites as being incompatible? I guess when there's at least one node not matching?

dom_test.py
```diff
@@ -0,0 +1,177 @@
...
+
+    l1 = list(a.iter(tag=etree.Element))
+    l2 = list(b.iter(tag=etree.Element))
+    s1 = []
+    s2 = []
+
+    print('Chrome Nodes : %d' % len(list(a.iter(tag=etree.Element))))
+    print('Firefox Nodes : %d' % len(list(b.iter(tag=etree.Element))))
+
+    if len(list(a.iter(tag=etree.Element))) + len(list(b.iter(tag=etree.Element))) > 1700:
```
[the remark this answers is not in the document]
marco-c: We don't care about speed at this point, we can make optimizations in the future.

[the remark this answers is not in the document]
marco-c: You can keep this check for now, but let's add a comment if x-pert doesn't do the same.

sagarvijaygupta:
@marco they don't give a solid label to a webpage as `incompatible`. They generate a report of `inconsistencies`. It might be there that screenshots vary at small point which makes `y_pred > 0.5` and hence marked as incompatible though our model would have found that small inconsistency.

dom_test.py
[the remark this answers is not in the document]
sagarvijaygupta: Okay!

sagarvijaygupta:
@marco-c Should I implement the next part in the same PR? In what form do we need the output? (I guess the same format we have written the labels.csv)

sagarvijaygupta:
@marco-c I will just write the results in form of CSV.

dom_test.py
```diff
@@ -0,0 +1,199 @@
...
+ignoredAttrib = {'style', 'type'}
+matched21 = {}
+matched12 = {}
+nodes_info = {1: {}, 2: {}}
+chrome_tree = None
+firefox_tree = None
+THRESHOLD_LEVEL = 0.75
+THRESHOLD_GLOBAL = 0.85
+folder = 'data'
+dom_files = [f for f in os.listdir(folder) if 'dom' in f and 'chrome' in f]
```
marco-c: Also add a `"firefox" in f`, to avoid the assumption that Firefox is always the first in the crawler.

sagarvijaygupta: So I will take intersection of dom_files_chrome and dom_files_firefox

dom_test.py
```diff
@@ -0,0 +1,200 @@
...
+matched12 = {}
+nodes_info = {1: {}, 2: {}}
+chrome_tree = None
+firefox_tree = None
+THRESHOLD_LEVEL = 0.75
+THRESHOLD_GLOBAL = 0.85
+folder = 'data'
+dom_files_chrome = ['_'.join(f.split('_')[:-1]) for f in os.listdir(folder) if 'dom' in f and 'chrome' in f]
+dom_files_firefox = ['_'.join(f.split('_')[:-1]) for f in os.listdir(folder) if 'dom' in f and 'firefox' in f]
+dom_files = [f for f in dom_files_chrome if f in dom_files_firefox]
```
marco-c: `dom_files = list(set(dom_files_chrome) + set(dom_files_firefox))`

sagarvijaygupta: I think you meant:
`dom_files = list(set(dom_files_chrome) & set(dom_files_firefox))`

marco-c: Yep, the intersection.

PR #228, opened 2018-07-05
Add complete screenshot for every webpage.

sagarvijaygupta:
Screenshots naming: BUGID_screenshotNumber_H_hshift_V_vshift_BROWSER.png


sagarvijaygupta:
I think we should also modify `label.py` so that screenshots of same page come together. I think it is needed?

collect.py
```diff
@@ -200,9 +200,21 @@ def do_something(driver, elem_properties=None):
 
 
 def screenshot(driver, file_path):
-    driver.get_screenshot_as_file(file_path)
-    image = Image.open(file_path)
-    image.save(file_path)
+    WINDOW_HEIGHT = 732
+    WINDOW_WIDTH = 412
```
marco-c: The size of the window is not the same on Firefox and Chrome, maybe we should make it the same since now we are removing the scrollbars.
Can you try to make the size the same and collect an example screenshot and post it here?

sagarvijaygupta: @marco-c I changed size of firefox window from 808 to 732. Size of screenshot changes.
I tested it on previous (808) and it works fine.

marco-c: What are the sizes of the screenshots for Firefox and Chrome before and after modifying the window size?

sagarvijaygupta:
These are on initial (different sizes)
1526_2_h_0_v_3660_chrome
1526_2_h_0_v_3660_firefox

These are after changing to 732 (making same)
8887_0_h_0_v_0_chrome

collect.py
sagarvijaygupta: FIREFOX : 412 x 656
CHROME : 412 x 732

marco-c:
> These are after changing to 732 (making same)

The second image was not uploaded :)

collect.py
marco-c: > FIREFOX : 412 x 656
> CHROME : 412 x 732

This is before or after? (I guess after?)

sagarvijaygupta:
After changing to 732 (making both same)
8887_0_h_0_v_0_chrome
8887_0_h_0_v_0_firefox

Inititally (with different size)
1526_1_h_0_v_2928_chrome
1526_1_h_0_v_2928_firefox

Sorry, didn't realise that!

collect.py
sagarvijaygupta: Yeah. Before they both were same and results were fine. So scrolling was same in both (if they starting at same position)

marco-c:
OK, so I guess we can't make them the same unfortunately :(
Or maybe we should do it and resize the images later.
Let's file an issue to figure this out, and in this PR take into account that the window sizes can be different.

sagarvijaygupta:
I will open the issue.

collect.py
marco-c: Oh, so the inner window size is the same? So we don't need to handle different window sizes here?

sagarvijaygupta: Yes! They are same. I matched their sizes and how do they scroll for different webpages.

collect.py
```diff
@@ -200,9 +200,21 @@ def do_something(driver, elem_properties=None):
...
+    page_height = driver.execute_script('return document.body.scrollHeight;')
+    page_width = driver.execute_script('return document.body.scrollWidth;')
+    height = 0
+    while height < page_height:
+        width = 0
+        while width < page_width:
+            file_name = '_'.join(file_path.split('_')[:-1] + ['H', str(width), 'V', str(height)] + file_path.split('_')[-1:])
```
marco-c: I'm not particulary fond of this, but I'll accept it for now.
We should add some code in utils to build a name and to parse a name.

sagarvijaygupta: Should I create a new PR with the change?

marco-c: Let's file an issue, it's not a high priority.

PR #231, opened 2018-07-05
Add functionality for images of same webpage to come together.

sagarvijaygupta:


sagarvijaygupta:
@marco-c This PR is ready. All screenshots of a given bugId come in sequence now.

label.py
```diff
@@ -359,8 +360,29 @@ def show_help():
...
+        if f1.isdigit():
+            return int(f1) - int(f2)
+
+
 def main():
+    global images_to_show
     show_help()
+
+    images_to_show = sorted(images_to_show, key=functools.cmp_to_key(images_cmp))
+    bug_ids = set([file_name.split('_')[0] for file_name in images_to_show])
```
marco-c: Another case where it would be useful to have the parsing function in utils :)

label.py
```diff
@@ -359,8 +360,29 @@ def show_help():
...
+    ordered_images = []
+    for bug_id in bug_ids:
+        ordered_images.extend([file_name for file_name in images_to_show if bug_id == file_name.split('_')[0]])
```
marco-c: You can use a list comprehension here instead, with two for loops in the list comprehension

label.py
```diff
@@ -359,8 +360,29 @@ def show_help():
...
+            continue
+        if f1.isdigit():
+            return int(f1) - int(f2)
+
+
 def main():
+    global images_to_show
     show_help()
+
+    images_to_show = sorted(images_to_show, key=functools.cmp_to_key(images_cmp))
```
marco-c: Let's do this stuff in a function, maybe we can call it `group_images`

label.py
sagarvijaygupta: :smile: Yeah! It feels a needful and will be easier to understand. I will add it after this PR is merged, changing it in both collect.py and here.

label.py
```diff
@@ -359,8 +360,34 @@ def show_help():
...
+        return len(img_list_x) - len(img_list_y)
+    for (f1, f2) in zip(img_list_x, img_list_y):
+        if f1 == f2:
+            continue
+        if f1.isdigit():
+            return int(f1) - int(f2)
+
+
+def group_images():
+    sorted_images_to_show = sorted(images_to_show, key=functools.cmp_to_key(images_cmp))
```
marco-c: Change the images_to_show in this function directly, instead of doing it in main

label.py
```diff
@@ -359,8 +360,34 @@ def show_help():
...
+    bug_ids = set([file_name.split('_')[0] for file_name in sorted_images_to_show])
+    ordered_images = []
+    for bug_id in bug_ids:
+        for file_name in sorted_images_to_show:
+            if bug_id == file_name.split('_')[0]:
+                ordered_images.append(file_name)
```
marco-c: Better to use a list comprehension here, similar to what you were doing before, but with two for loops in the comprehension rather than one.

marco-c: ordered_images = [XXX for bug_id in bug_ids for file_name in sorted_images_to_show if bug_id == ...]

PR #234, opened 2018-07-16
Shane labelling

sdv4:
Will be updated with new labels as labelling is completed


label_persons/data
```diff
@@ -0,0 +1,3 @@
+Image Name,Label
```
marco-c: This file probably should be removed?

marco-c:
Instead of updating this PR, let's merge it (once you remove the unneeded files), and then you can submit additional PRs when you have more labels.

label_persons/data
marco-c: I didn't mean the header row, that row in label_persons/shane.csv should be kept, I meant this file (label_persons/data).

marco-c:
Could you rebase on top of master? I've fixed https://github.com/marco-c/autowebcompat/issues/240, which means the diff will be smaller now.

The easiest way to rebase would be to:
1) remove the changes from labels.csv;
2) run the label.py script (and press ESC right away) to sort your labels;
3) run the generate_labels.py script to merge your labels with the other labels.

Then update this PR and I'll merge it!

sdv4:
@marco-c when you say "remove changes from labels.csv", are you referring to my previous removal of the 'Image name, label' header column?

autowebcompat/utils.py
```diff
@@ -192,7 +192,7 @@ def read_labels(file_name='labels.csv'):
         with open(file_name, 'r') as f:
             next(f)
             reader = csv.reader(f)
-            labels = {row[0]: row[1] for row in reader}
+            labels = {row[0]: row[1] for row in reader if len(row) == 2}
```
marco-c: This check should no longer be needed on latest master (unless you're generating labels on Windows, which sometimes can cause problems. Are you?)

marco-c:
For some reason your labels are not sorted :S 
Did you run label.py after rebasing on top of master?

marco-c:
@sdv4 do you want me to try to update the PR? Or are you going to do it?

sdv4:
Sorry, I will do it ASAP!

autowebcompat/utils.py
sdv4: I was having this problem on MacOS and Ubuntu, but the problem went away with the check.

marco-c: OK, can you try again now with latest master? It should be fixed

marco-c: @sdv4 ?

marco-c:
There are still some conflicts to resolve (otherwise merging is blocked).
The easiest way to resolve them is:
1) Remove changes from labels.csv
2) Pull from master
3) Run generate_labels.py

marco-c:
Done in #255 (commit attributed to Shane).

sdv4:
@marco-c thanks for merging the labels in #255. I just checked and the extra conditional is no longer needed in utils.py when running from Ubuntu.

PR #237, opened 2018-07-20
Prefer images that haven't been labeled by anyone yet.

### LINKED ISSUE #235: Label script should prefer images that haven't been labeled by anyone yet
This way we can more quickly get a larger number of labels.

sagarvijaygupta:
Fixes #235.

### DIFF
--- a/label.py
+++ b/label.py
@@ -1,4 +1,5 @@
 import argparse
+from collections import OrderedDict
 import functools
 import random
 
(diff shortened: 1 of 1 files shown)


label.py
```diff
@@ -20,8 +21,12 @@
 if args.verify:
     images_to_show = [i for i in utils.get_images() if i in labels]
 else:
-    images_to_show = [i for i in utils.get_images() if i not in labels]
-    random.shuffle(images_to_show)
+    all_labels = utils.read_labels()
+    unlabelled_in_all_labels = [i for i in utils.get_images() if i in all_labels and i not in labels]
```
marco-c: Use images_to_show here and for `unlabelled`, so you don't have to call get_images() twice.

label.py
```diff
@@ -375,7 +380,7 @@ def images_cmp(x, y):
 def group_images():
     global images_to_show
     sorted_images_to_show = sorted(images_to_show, key=functools.cmp_to_key(images_cmp))
-    bug_ids = set([file_name.split('_')[0] for file_name in sorted_images_to_show])
+    bug_ids = OrderedDict.fromkeys([file_name.split('_')[0] for file_name in unlabelled + unlabelled_in_all_labels])
```
marco-c: We want to show the `unlabelled_in_all_labels` first

sagarvijaygupta: It was a naming confusion. `unlabelled_in_all_labels` actually meant `unlabelled_but_present_in_all_labels`

marco-c: Ahahah, yes, thanks!

Issue #238, opened 2018-07-20
Alignment Graph in xpert


sagarvijaygupta:
As mentioned in #226 

> In the AlignmentGraph they check for the matched nodes:
> 1. They create an AlignmentGraph for two doms where they create a vMap (mapping from xPath to node), then sort the vertices based on Area and DOM hierarchy, and add nodes as a pair of (parent, node) when parent contains the node (node is inside the parent visually and considering large number of factors). Similarly add siblings who have same parent.
> 2.  For all pair of matched nodes they compare their parents and siblings for large variety of criteria including different alignments, different parents, vfill, hfill etc considering many size differences greater than some threshold for both parents and siblings. These return large number of issues to the checker.

Second part of matching is Alignment Graph.

sagarvijaygupta:
>  First part of this should be having the coordinates of bounding box for every element on the webpage.

@marco-c this part I think it should be implemented in Crawler itself. I will output the result into a json format from the Crawler.

PR #239, opened 2018-07-20
Write locations of all elements on a webpage.

sagarvijaygupta:
Solves #238 partially


sagarvijaygupta:
@marco-c This will slow down the crawler. Implementing it separately may lead to inconsistency as in that case, what xpert will do won't be at the same time when we have screenshots for.

collect.py
```diff
@@ -12,8 +12,10 @@
 from selenium.common.exceptions import NoAlertPresentException
 from selenium.common.exceptions import NoSuchWindowException
 from selenium.common.exceptions import TimeoutException
+from selenium.common.exceptions import NoSuchElementException
 
 from autowebcompat import utils
+from lxml import etree
```
sagarvijaygupta: Moving it in above code segment (where it should be to pass travis) leads to some problem. Only screenshots get captured, dom information is not written, traversing fails. I am not getting how to resolve it. It might be overwriting some previous import but i am not sure how to handle it.

collect.py
```diff
@@ -222,10 +224,27 @@ def get_domtree(driver, file_path):
...
 
 
+def get_coordinates(driver, file_path):
+    dom_tree = etree.HTML(driver.execute_script('return document.documentElement.outerHTML'))
+    dom_element_tree = etree.ElementTree(dom_tree)
+    loc_dict = {}
+    for elem in dom_tree.iter(tag=etree.Element):
+        xpath = dom_element_tree.getpath(elem)
+        try:
+            web_elem = driver.find_element_by_xpath(xpath)
```
marco-c: We do we need this?

collect.py
marco-c: Maybe try renaming the import, e.g. `from lxml import etree as lxml_etree`

collect.py
sagarvijaygupta: I know all xpaths should be there. But while testing I did find some cases where driver couldn't find the element. So I had to put this, otherwise it would just leave all other elements.

collect.py
sagarvijaygupta: I did try this. It still didn't work. I thought in this way it won't over-write.  :disappointed:

marco-c: Really strange, I'll try to reproduce locally.

collect.py
marco-c: How often does it happen? Can you check with a few websites (say 100)?
Maybe we can implement it in another way and not use lxml here.

sagarvijaygupta: Its not that often! Even if it is, it is generally for `Scalable Vector Graphics` or `noscript` and that is less than 0.1%. More than 50% of these locations will not ignored in xperts implementation. I will provide with experimental stats in a while.

sagarvijaygupta: @marco-c we can implement it in another way, but we will need a dictionary of the form {xpath : size, location}. Since lxml is used for parsing the DOM in xpert xpaths should be same.

sagarvijaygupta: If we want to have all elements I feel there is only one possible way, ie for all elements found in selenium we create the xpath by ourselves.

sagarvijaygupta: I implemented getting xpath from the selenium web element, this is a significant improvement in speed in comparison to present approach. 
1. While Implementing this I made the xpaths same as given by lxml, problem with `Scalable Vector Graphics` is solved.
2.  `no script` element is only present in lxml dom tree and no element exists corresponding to it in selenium elements acquired by `driver.find_elements_by_css_selector('*')`
3. There are some elements which have which have a `div`in dom's xpath while in xpath got from element we get `div'`
4. There are some elements which have `body` in dom's xpath while `head` in xpath got from element.
5. In some cases there are more number of elements found using element xpath than xpath got from dom. (strange)

Presently in 33 test cases 31 matched perfectly.
1 has 4th issue
1 has 5th issue (which is useless because we will be comparing elements we get from lxml dom)

sagarvijaygupta: [out.txt](https://github.com/marco-c/autowebcompat/files/2216407/out.txt)

Error is actually very less as xpert ignores small children. 
If you feel this error is tolerable I will continue to work on it. Else I will have to find something different.

sagarvijaygupta: Number of mismatch 13
Number of chrome 42
Number of firefox 60
Total Number 102
From above text file

marco-c: I think the error is tolerable! Let me review the current code

collect.py
```diff
@@ -222,10 +226,120 @@ def get_domtree(driver, file_path):
...
+    loc_dict = {}
+    dom_tree_elements = [elem for elem in dom_tree.iter(tag=etree.Element)]
+    web_elements = driver.find_elements_by_css_selector('*')
+    left_elements = []
+    dom_xpaths = []
+
+    for element in dom_tree_elements:
+        xpath = dom_element_tree.getpath(element)
+        dom_xpaths.append(xpath)
+    for element in web_elements:
```
marco-c: Nit: newline before this for

collect.py
```diff
@@ -222,10 +226,120 @@ def get_domtree(driver, file_path):
         f.write(driver.execute_script('return document.documentElement.outerHTML'))
 
 
+def get_coordinates(driver, file_path):
+    dom_tree = etree.HTML(driver.execute_script('return document.documentElement.outerHTML'))
+    dom_element_tree = etree.ElementTree(dom_tree)
+    loc_dict = {}
+    dom_tree_elements = [elem for elem in dom_tree.iter(tag=etree.Element)]
+    web_elements = driver.find_elements_by_css_selector('*')
```
marco-c: Can you use the same find_elements we use in the part of the crawler where we search for all elements on the page to click on them?

collect.py
```diff
@@ -222,10 +226,120 @@ def get_domtree(driver, file_path):
...
+    left_elements = []
+    dom_xpaths = []
+
+    for element in dom_tree_elements:
+        xpath = dom_element_tree.getpath(element)
+        dom_xpaths.append(xpath)
+    for element in web_elements:
+        xpath = driver.execute_script("""function absoluteXPath(element) {
```
marco-c: Since this is a huge script, I think we should write it in a separate JS file and load this JS file at the startup of the crawler in a variable, and then use the variable here.

marco-c: Did you find this implementation online or did you implement it from scratch?

collect.py
```diff
@@ -222,10 +226,120 @@ def get_domtree(driver, file_path):
...
+                                    }
+                                    return absoluteXPath(arguments[0]);""", element)
+
+        if xpath in dom_xpaths:
+            loc_dict[xpath] = element.size
+            loc_dict[xpath].update(element.location)
+            dom_xpaths.remove(xpath)
+
+    for xpath in dom_xpaths:
+            try:
```
marco-c: There's some wrong indentation here

marco-c:
I think this looks good (after you remove the debugging code)! The number of mismatches is reasonably small.

collect.py
sagarvijaygupta: >  buttons = body.find_elements_by_tag_name('button')
    selects = body.find_elements_by_tag_name('select')
    children = buttons + links + inputs + selects

I think you are talking about this? But we want all elements and not just these.

collect.py
sagarvijaygupta: I found the script online to which I had to make some changes so that the xpath generated by the script and that we find using lxml come out to be exactly the same. Otherwise they both won't be compatible for future part of xpert.

collect.py
marco-c: I thought we were also getting all elements somewhere, but my memory failed me :P

collect.py
marco-c: OK! Where did you find it? We should make sure the license is compatible, or if you found it on stackoverflow you can add a comment at the beginning of the file with a link to the stackoverflow answer.

sagarvijaygupta: https://stackoverflow.com/questions/47069382/want-to-retrieve-xpath-of-given-webelement

I will add it as a comment in the beginning of the file.

sagarvijaygupta:
@marco-c Since we require this location part only for xpert and I don't think we will be running xpert for all the websites for which we will be running our models, do you feel that we need location files for all? Or should we just create a test folder where we will be having location files for few websites only.

collect.py
sagarvijaygupta: @marco-c were you able to reproduce the error locally?

sagarvijaygupta: @marco-c it was lxml version issue. It works fine with 4.2.3 (which was earlier 4.2.2)

requirements.txt
```diff
@@ -4,4 +4,5 @@ keras==2.1.1
 tensorflow==1.3.0
 h5py==2.7.1
 opencv-python==3.4.0.12
-scikit-learn==0.19.1 
\ No newline at end of file
+scikit-learn==0.19.1 
+lxml==4.2.3
```
sagarvijaygupta: Should I change 
`lxml==4.2.3` to `lxml!=4.2.2`

marco-c: No, pinning to 4.2.3 is better, this way we're sure all users get the same version we are using.

get_xpath.js
```diff
@@ -0,0 +1,71 @@
+// Taken with some modifications from - https://stackoverflow.com/questions/47069382/want-to-retrieve-xpath-of-given-webelement
+function absoluteXPath(element) {
+    var comp, comps = [];
```
marco-c: Use two character indentation for JS

get_xpath.js
```diff
@@ -0,0 +1,71 @@
...
+    var parent = null;
+    var xpath = '';
+    var getPos = function(element) {
+    var position = 0,
```
marco-c: Here and in all other cases, use `let` insead of `var`

collect.py
```diff
@@ -7,9 +7,11 @@
 import time
 import traceback
 
+from lxml import etree
```
marco-c: Moving this after `from PIL import Image` fixes the flake8 failure

collect.py
```diff
@@ -32,6 +34,7 @@
 
 bugs = utils.get_bugs()
 print(len(bugs))
+get_xpath_script = ''.join(open('get_xpath.js', 'r').readlines()[1:])
```
marco-c: Better to use `with`. Also, I think there's no need to skip the comment.
```Python
with open('get_xpath.js', 'r') as f:
    get_xpath_script = f.read()
```

get_xpath.js
```diff
@@ -0,0 +1,71 @@
...
+    curNode, prevNode, nextNode;
+    if (element.nodeType == Node.ATTRIBUTE_NODE) {
+        return null;
+    }
+    prevNode = element.previousSibling;
+    nextNode = element.nextSibling;
+    while(prevNode) {
```
marco-c: Space after `while`, `if` and so on (`while (prevNode)`)

get_xpath.js
marco-c: There are some problems with indentation in parts of the file (e.g. the indentation doesn't increase after `var getPos = function(element) {`).

collect.py
```diff
@@ -33,6 +35,8 @@
 bugs = utils.get_bugs()
 print(len(bugs))
 
+with open('get_xpath.js', 'r') as f:
+    get_xpath_script = f.read()
```
marco-c: Travis is failing as flake8 is expecting two blank lines before a function definition.

get_xpath.js
```diff
@@ -0,0 +1,70 @@
+// Taken with some modifications from - https://stackoverflow.com/questions/47069382/want-to-retrieve-xpath-of-given-webelement
+function absoluteXPath(element) {
+  var comp, comps = [];
```
marco-c: Nit: Use `let` instead of `var`

sagarvijaygupta: I did make that change :confused:

PR #242, opened 2018-07-24
Add name creation and parsing in utils.py

### LINKED ISSUE #232: Add screenshot name creation and parsing in utils.py
As discussed in #228

sagarvijaygupta:
Fixes #232.


autowebcompat/utils.py
```diff
@@ -293,3 +293,34 @@ def write_train_info(information, model, train_history, file_name=None):
...
+
+    new_file_name = file_info + '_H_' + str(width) + '_V_' + str(height) + '_' + browser + '.' + extension
+    return os.path.join(folder, new_file_name)
+
+
+def parse_file_name(file_name):
+    file_name = file_name.split('_')
+    file_info = {}
+    index = 0
+    file_info['bug_id'] = int(file_name[index])
```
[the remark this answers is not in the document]
marco-c: I forgot that in some cases we have a different number of elements, so keep using the index.

autowebcompat/utils.py
```diff
@@ -293,3 +293,34 @@ def write_train_info(information, model, train_history, file_name=None):
...
+    extension = file.split('.')[1]
+    file_info = file_name.rsplit('_', 1)[0]
+    browser = file_name.rsplit('_', 1)[1]
+
+    new_file_name = file_info + '_H_' + str(width) + '_V_' + str(height) + '_' + browser + '.' + extension
+    return os.path.join(folder, new_file_name)
+
+
+def parse_file_name(file_name):
+    file_name = file_name.split('_')
```
marco-c: Rename `file_name`, as it's not a great name. Maybe `file_name_parts`?

autowebcompat/utils.py
```diff
@@ -293,3 +293,34 @@ def write_train_info(information, model, train_history, file_name=None):
...
+def parse_file_name(file_name):
+    file_name_parts = file_name.split('_')
+    file_info = {}
+    file_info['bug_id'] = int(file_name_parts[0])
+
+    shift = 0
+    if len(file_name_parts) >= 2 and file_name_parts[1].isdigit():
+        file_info['seq_no'] = int(file_name_parts[1])
+        shift = 1
+    if len(file_name_parts) > 2:
```
sagarvijaygupta: Conditions are required since previous screenshots have different naming conventions

autowebcompat/utils.py
```diff
@@ -293,3 +293,34 @@ def write_train_info(information, model, train_history, file_name=None):
             row = train_history_list[i]
             for col in row:
                 print('%f\t\t' % col, end=' ', file=f)
+
+
+def create_screenshot_name(file_name_info):
```
marco-c: You can pass several arguments here instead of an object, it's cleaner

marco-c: seq_no can have the default of None

sagarvijaygupta: Should I just pass them differently? Can I store them as dictionary in collect.py?

autowebcompat/utils.py
```diff
@@ -293,3 +293,34 @@ def write_train_info(information, model, train_history, file_name=None):
...
+    if len(file_name_parts) >= 2 and file_name_parts[1].isdigit():
+        file_info['seq_no'] = int(file_name_parts[1])
+        shift = 1
+    if len(file_name_parts) > 2:
+        file_info['width'] = int(file_name_parts[shift + 2])
+        file_info['height'] = int(file_name_parts[shift + 4])
+    return file_info
+
+
+def create_dom_name(file_name_info):
```
marco-c: This function should reuse create_screenshot_name, since the format is the same except for the dom_ part

autowebcompat/utils.py
marco-c: And let's call the function `create_file_name`, to match the equivalent but opposite `parse_file_name`.

marco-c: > Should I just pass them differently? Can I store them as dictionary in collect.py?

What do you mean?

sagarvijaygupta: I am creating an object like this
> image_file_info['folder'] = 'data'
image_file_info['bug_id'] = str(bug['id'])
 image_file_info['browser'] = browser

in collect.py 
Should I keep it this way only? And when calling the above function just call them like
> create_screenshot_name(image_file_info['folder'],  image_file_info['bug_id'],  image_file_info['width'],  image_file_info['height'])

Or should I keep different variables for them?

marco-c: Keep different variables, the argument list will get longer (but not too long) but you won't needlessly create an object

autowebcompat/utils.py
sagarvijaygupta: But create_screenshot_name uses height and width also as parameters which is not for dom

marco-c: You can make them `None` by default and not add them when creating the name for dom

sagarvijaygupta: Yeah! But it will result in some if else in create_file_name. I hope it will be fine in that case?

marco-c: Sure!

autowebcompat/utils.py
```diff
@@ -293,3 +293,39 @@ def write_train_info(information, model, train_history, file_name=None):
             row = train_history_list[i]
             for col in row:
                 print('%f\t\t' % col, end=' ', file=f)
+
+
+def create_file_name(folder, bug_id, browser, width=None, height=None, seq_no=None, isDom=False):
+    new_file_name_parts = []
+    if isDom:
```
marco-c: Remove this and just do it in crawler

autowebcompat/utils.py
```diff
@@ -293,3 +293,39 @@ def write_train_info(information, model, train_history, file_name=None):
...
+        new_file_name_parts.append('dom')
+    new_file_name_parts.append(bug_id)
+    if seq_no is not None:
+        new_file_name_parts.append(seq_no)
+    if width is not None:
```
marco-c: Nit: newline before ifs

autowebcompat/utils.py
```diff
@@ -293,3 +293,39 @@ def write_train_info(information, model, train_history, file_name=None):
...
+        new_file_name_parts.append('H')
+        new_file_name_parts.append(width)
+    if height is not None:
+        new_file_name_parts.append('V')
+        new_file_name_parts.append(height)
+    new_file_name_parts.append(browser)
+    new_file_name = '_'.join(new_file_name_parts)
```
marco-c: Return this, then in crawler you'll add the necessary extension

collect.py
```diff
@@ -217,15 +217,15 @@ def screenshot(driver, file_path):
...
 
-def get_domtree(driver, file_path):
-    with open(file_path, 'w', encoding='utf-8') as f:
+def get_domtree(driver, folder, bug_id, browser, seq_no):
+    with open(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True), 'w', encoding='utf-8') as f:
         f.write(driver.execute_script('return document.documentElement.outerHTML'))
 
 
-def get_screenshot_and_domtree(driver, file_name):
+def get_screenshot_and_domtree(driver, folder, bug_id, browser, seq_no=None):
```
marco-c: Remove the `folder` argument, as it's always 'data'

tests/test_utils.py
```diff
@@ -135,3 +135,44 @@ def test_to_categorical_label():
...
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, width=width, height=height, seq_no=seq_no) == 'test/1661_2_H_20_V_10_chrome.png')
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True) == 'test/dom_1661_2_chrome.txt')
+
+
+def test_parse_file_name():
+    file_name = '1661'
+    utils_file_name_info = utils.parse_file_name(file_name)
+    file_name_info = {}
+    file_name_info['bug_id'] = 1661
+    assert(file_name_info == utils_file_name_info)
```
marco-c: assert file_name_info == utils.parse_file_name(file_name)

tests/test_utils.py
```diff
@@ -135,3 +135,44 @@ def test_to_categorical_label():
...
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True) == 'test/dom_1661_chrome.txt')
+    seq_no = '2'
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, width=width, height=height, seq_no=seq_no) == 'test/1661_2_H_20_V_10_chrome.png')
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True) == 'test/dom_1661_2_chrome.txt')
+
+
+def test_parse_file_name():
+    file_name = '1661'
+    utils_file_name_info = utils.parse_file_name(file_name)
+    file_name_info = {}
```
marco-c:
```
info = {
    'bug_id': 1661,
}
```

tests/test_utils.py
```diff
@@ -135,3 +135,44 @@ def test_to_categorical_label():
...
+    width = '20'
+    browser = 'chrome'
+    folder = 'test'
+    seq_no = None
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, width=width, height=height, seq_no=seq_no) == 'test/1661_H_20_V_10_chrome.png')
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True) == 'test/dom_1661_chrome.txt')
+    seq_no = '2'
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, width=width, height=height, seq_no=seq_no) == 'test/1661_2_H_20_V_10_chrome.png')
+    assert(utils.create_file_name(folder=folder, bug_id=bug_id, browser=browser, seq_no=seq_no, isDom=True) == 'test/dom_1661_2_chrome.txt')
+
```
marco-c: Also assert the case in which you have no height and width

tests/test_utils.py
```diff
@@ -135,3 +135,44 @@ def test_to_categorical_label():
...
     assert categorical_label == 1
+
+
+def test_create_file_name():
+    bug_id = '1661'
+    height = '10'
+    width = '20'
+    browser = 'chrome'
+    folder = 'test'
+    seq_no = None
```
marco-c: Hardcode these into the call directly instead of defining them as variables

PR #249, opened 2018-07-26
Fix newline issue on Windows

amaaniqbal:
This pull request fixes extra generated newline on Windows while writing to csv file. The generated newline was causing error while reading from csv file.


marco-c:
Are you sure this is a problem?
For reading, we are opening the file with `newline=None`, so it means universal newline mode is enabled, which means all newlines are translated to `\n`.

amaaniqbal:
I guess so because for reading if there is an extra newline then error is being thrown and that extra newline is generated while writing to csv file.

Atleast this has been the case with me and doing above step solved the issue on Windows. You can browse through [](https://stackoverflow.com/questions/3348460/csv-file-written-with-python-has-blank-lines-between-each-row/3348664) for details.

Its better to see its behavior on other OS too.

marco-c:
OK, makes sense. Please switch the csv reading to use newline='' too, as that's suggested by the Python documentation: https://docs.python.org/3/library/csv.html#id3.

PR #252, opened 2018-07-31
Migrate files according to new convention

### LINKED ISSUE #251: Script to rename images and labels according to new convention
As mentioned in #88, we need a script to `rename images` in `data`, `text files` corresponding to `bug_id` and `label_files` along with `bounding_box_files` in `label_persons`

sagarvijaygupta:
Migrate data image files, text files, labelled files in label_persons according to the new conventions incorporated in #88

Fixes #251


marco-c:
Let's put the code in a function `migrate_v1_to_v2`, as we might have other migrations in the future.
We should put a VERSION file in the data/ repository, so we know which migration should be applied.

sagarvijaygupta:
What should we write in VERSION file?

marco-c:
> What should we write in VERSION file?

Right now, just `1`. Then the migrate script should check the VERSION file, if it's `1` it should call `migrate_v1_to_v2` (which changes it to `2`).

marco-c:
(The VERSION file should be added to the autowebcompat-data repo on GitLab)

sagarvijaygupta:
> Then the migrate script should check the VERSION file, if it's 1 it should call migrate_v1_to_v2 (which changes it to 2)

You mean that after checking VERSION I should call `migrate_v1_to_v2` by default?
We expect future migrations of the form   `migrate_v2_to_v3`,  `migrate_3_to_v4`  or  `migrate_v1_to_v3`,  `migrate_v1_to_v4`?

sagarvijaygupta:
@marco-c should I make the `to_version` as argparse input?

marco-c:
> You mean that after checking VERSION I should call migrate_v1_to_v2 by default?
We expect future migrations of the form migrate_v2_to_v3, migrate_3_to_v4 or migrate_v1_to_v3, migrate_v1_to_v4?

`migrate_v1_to_v2` shouldn't always be called, but only if VERSION is == `1`.
Future migrations will be `migrate_v2_to_v3`, `migrate_v3_to_v4` and so on, this way we can support all possible migrations.

marco-c:
> @marco-c should I make the to_version as argparse input?

The to_version should always be the latest (in this case 2).

sagarvijaygupta:
Okay. So should I recursively call all of them and update the VERSION file in the end or we will just call it once and if wanted change VERSION file by ourselves and again call?

marco-c:
> Okay. So should I recursively call all of them and update the VERSION file in the end or we will just call it once and if wanted change VERSION file by ourselves and again call?

Not recursively, but one after another. Each function should also update the VERSION file.

In pseudocode, something like this:
```
for i in range(CURRENT_VERSION, LATEST_VERSION - 1):
    migrate_v(i)_to_v(i+1)
```

sagarvijaygupta:
@marco-c Can you see the VERSION file now? I added it previously also. I realized I had to commit in the submodule also before adding.

marco-c:
You can't update the submodule without actually updating it on GitLab, otherwise it's not available to anyone else except you (infact, Travis fails to clone the submodule now, as the hash you provide is local to your machine).

Don't worry about adding the VERSION file, as I'm doing it now.

For the future, when you want to change something in data or tools, you actually have to send a pull request on GitLab, so that it can be merged and can be available to everybody.

sagarvijaygupta:
Okay! I will try it next time. I actually wanted to update some new full_page_screenshots previously but was not able to at that time!

sagarvijaygupta:
Should we keep a file along with VERSION file which corresponds to the screenshots present in that version. Since we won't be able to push images if generated from present crawler as after that migrate function will try to migrate them also.

marco-c:
> Should we keep a file along with VERSION file which corresponds to the screenshots present in that version. Since we won't be able to push images if generated from present crawler as after that migrate function will try to migrate them also. (or simply the time stamp?)

I guess it's better if we just do the migration before adding new images.

migrate_files.py
```diff
@@ -0,0 +1,95 @@
+import os
+
+from autowebcompat import utils
+
+LATEST_VERSION = 5
```
marco-c: LATEST_VERSION should be 2

migrate_files.py
```diff
@@ -0,0 +1,95 @@
...
+        utils.write_bounding_boxes(new_bounding_boxes, os.path.join('label_persons', f))
+
+
+with open(os.path.join('data', 'VERSION'), 'r') as f:
+    CURRENT_VERSION = int(f.read())
+
+while CURRENT_VERSION != LATEST_VERSION:
+    migrate_function = globals()['migrate_v%d_to_v%d' % (CURRENT_VERSION, CURRENT_VERSION + 1)]
+    migrate_function()
+    CURRENT_VERSION += 1
```
marco-c: We also need to write the new version in the VERSION file

migrate_files.py
```diff
@@ -0,0 +1,98 @@
+import os
+
+from autowebcompat import utils
+
+LATEST_VERSION = 2
+
+
+def read_sequence(bug_id):
```
marco-c: This function and the next should be defined under `migrate_v1_to_v2`, as they are specific to this migration (if there's going to be a v2 to v3 in the future, these two functions will not be usable in both migrations).

migrate_files.py
```diff
@@ -0,0 +1,98 @@
...
+            continue
+        bug_id = os.path.splitext(f)[0]
+        data = read_sequence(int(bug_id))
+        write_sequence(int(bug_id), data)
+
+    for f in label_files:
+        if 'csv' not in f:
+            continue
+        labels = utils.read_labels(os.path.join('label_persons', f))
+        new_labels = {}
```
marco-c: You can directly modify `labels` instead.

migrate_files.py
```diff
@@ -0,0 +1,98 @@
...
+
+        for key, value in labels.items():
+            key_info = utils.parse_file_name(key)
+
+            if 'seq_no' not in key_info:
+                new_labels[key] = value
```
marco-c: if you directly modify labels instead, you won't need to do this

migrate_files.py
marco-c: Actually, it's clearer this way, let's keep it this way.

migrate_files.py
```diff
@@ -0,0 +1,98 @@
...
+            browser = parts[2]
+            seq_no = map_names[seq_no]
+
+            key = utils.create_file_name(bug_id=bug_id, browser=browser, seq_no=seq_no)
+            new_bounding_boxes[key] = value
+        utils.write_bounding_boxes(new_bounding_boxes, os.path.join('label_persons', f))
+
+
+with open(os.path.join('data', 'VERSION'), 'r') as f:
+    CURRENT_VERSION = int(f.read())
```
marco-c: Nit: Usually upper case is for constants, let's use lower case here since you modify this later.

PR #253, opened 2018-07-31
Use xpath to jump_back

sagarvijaygupta:


marco-c:
Could you rebase on top of master? The history is quite confusing

collect.py
```diff
@@ -304,12 +308,12 @@ def count_lines(bug_id):
 
 
 # We restart from the start and follow the saved sequence
-def jump_back(current_path, firefox_driver, chrome_driver, visited_paths, bug):
+def jump_back(current_path, current_path_info, firefox_driver, chrome_driver, visited_paths, bug):
```
marco-c: Let's use `current_path_node` or `current_path_elem`, it makes it more clear that it is a node or element in a current path list.

marco-c: Ah no, this is actually another representation of the path. I guess we need to refactor this a bit otherwise it's confusing.

marco-c: What about removing `current_path` and just using `current_path_info` always instead? (renaming `current_path_info` to `current_path`)

sagarvijaygupta: I will have to make some changes in previously defined visit and other functions. I will update the PR shortly.

sagarvijaygupta: @marco-c can I keep `current_path_info` as an object with structure like: 
current_path_info ={'elements' :  [ ], 'chrome_xpath': [ ], 'firefox_xpath': [ ]}
This way I will be able to keep visited_paths as a path of elements only.
If I keep `current_path_info` in the way it is presently (list with three columns same as above), I will have to create a list of elements to check if it exists in `visited_paths`. 
I will do it the way you prefer.

marco-c: The current format looks better to me, instead of three lists.
You can just do `[element for element, firefox_xpath, chrome_xpath in current_path_info]` to retrieve the path of elements.
If you have to do it too many times and the code gets too confusing, then we can do the approach with three lists, but if it's just a couple of times, then the approach with a single list and tuples looks better.

collect.py
```diff
@@ -156,14 +152,13 @@ def do_something(driver, visited_paths, current_path, elem_properties=None):
...
-            if was_visited(current_path, visited_paths, elem_properties):
-                continue
-
             # If the element is not displayed or is disabled, the user can't interact with it. Skip
             # non-displayed/disabled elements, since we're trying to mimic a real user.
             if not child.is_displayed() or not child.is_enabled():
-                current_path.pop()
+                continue
+
+            # we check if the path has been visited previously else visit it
```
marco-c: `# Check if the path has been visited previously, visit it if it hasn't`

The comment at line 166 doesn't apply anymore, since we visit now.

collect.py
```diff
@@ -172,19 +167,22 @@ def do_something(driver, visited_paths, current_path, elem_properties=None):
...
             else:
                 children_to_ignore.extend(elems)
-            visit(current_path, visited_paths)
-            break
     else:
-        if 'id' not in elem_properties['attributes'].keys():
+        if 'id' in elem_properties['attributes'].keys():
+            elem_id = elem_properties['attributes']['id']
+            elem = driver.find_element_by_id(elem_id)
+            xpath = driver.execute_script(get_xpath_script, elem)
```
marco-c: Ah this is going to slow down a bit the best case (elements with IDs) :(

collect.py
```diff
@@ -225,8 +223,7 @@ def do_something(driver, visited_paths, current_path, elem_properties=None):
                 break
 
     close_all_windows_except_first(driver)
-
```
marco-c: Don't remove this newline

collect.py
```diff
@@ -347,11 +344,13 @@ def run_test_both(bug, firefox_driver, chrome_driver):
...
 
-        do_something(chrome_driver, visited_paths, current_path, elem_properties)
+        _, chrome_xpath = do_something(chrome_driver, visited_paths, current_path, elem_properties)
+        current_path.append((elem_properties, firefox_xpath, chrome_xpath))
+
         with open('data/%d.txt' % bug['id'], 'a+') as f:
             line_number = count_lines(bug['id'])
             f.seek(2, 0)
-            for element in current_path:
+            for (element, _, _) in current_path:
```
marco-c: Nit: no parenthesis, you can just do `for element, _, _ in current_path`

collect.py
```diff
@@ -172,19 +167,22 @@ def do_something(driver, visited_paths, current_path, elem_properties=None):
...
-            visit(current_path, visited_paths)
-            break
     else:
-        if 'id' not in elem_properties['attributes'].keys():
+        if 'id' in elem_properties['attributes'].keys():
+            elem_id = elem_properties['attributes']['id']
+            elem = driver.find_element_by_id(elem_id)
+            xpath = driver.execute_script(get_xpath_script, elem)
+        elif xpath is not None:
+            elem = driver.find_element_by_xpath(xpath)
```
marco-c: If we can't find the element via xpath, we should fallback on finding it via the properties.

marco-c: Or maybe we should just assert that we found the element here.

collect.py
sagarvijaygupta: I will put a check to update xpath only if it is None

PR #256, opened 2018-08-02
Alignment Graph in xpert

sagarvijaygupta:


sagarvijaygupta:
@marco-c What is expected as output from the alignment graph? What xpert does is it creates a list of issues. I don't think we need that? So what should I provide as the output? 
Same as what we have decided for the basic part ie 'y' or 'n'? So a node will be considered compatible if all its attributes match 100% and even a slight issue with one node will consider the whole webpage incompatible. Right?

sagarvijaygupta:
```
// I will be matching diffY for both c1 and c2 (this is incorrect in xpert)
if(c1.isSizeDiffX() && c2.isSizeDiffY()) {
	if (testSizeDiff(c1.topAligned, c2.topAligned, c1.yError,c2.yError)) {
		issues.add(errMsg("TOP-ALIGNMENT", c1, c2));
	}

	if (testSizeDiff(c1.bottomAligned, c2.bottomAligned, c1.yError, c2.yError)) {
		issues.add(errMsg("BOTTOM-ALIGNMENT", c1, c2));
	}
	
	if(testSizeDiff(c1.middle, c2.middle, c1.yError, c2.yError)) {
		issues.add(errMsg("VMID-ALIGNMENT", c1, c2));
	}
	
	if(c1.vFill ^ c2.vFill) {
		issues.add(errMsg("VFILL", c1, c2));
	}
}

if(c1.isSizeDiffX() && c2.isSizeDiffX()) {
	if(c1.hFill ^ c2.hFill) {
		issues.add(errMsg("HFILL", c1, c2));
	}
	
	if (c1.centered ^ c2.centered) {
		issues.add(errMsg("CENTER-ALIGNMENT", c1, c2));
	}

	if (c1.leftJustified ^ c2.leftJustified) {
		issues.add(errMsg("LEFT-JUSTIFICATION", c1, c2));
	}

	if (c1.rightJustified ^ c2.rightJustified) {
		issues.add(errMsg("RIGHT-JUSTIFICATION", c1, c2));
	}
}
}	 
```
Just for the record as this thing will be different from original implementation

marco-c:
> @marco-c What is expected as output from the alignment graph? What xpert does is it creates a list of issues. I don't think we need that? So what should I provide as the output?
Same as what we have decided for the basic part ie 'y' or 'n'? So a node will be considered compatible if all its attributes match 100% and even a slight issue with one node will consider the whole webpage incompatible. Right?

Yes, then we can change the threeshold later if we want it to be less strict.

marco-c:
> Just for the record as this thing will be different from original implementation

Also add a comment in the code explaining that it is different and why.

marco-c:
> I feel if p1 and p2 are same it should return true else it should do the given stuff and then return corresponding result.

testSizeDiff should return true if there's a difference and false if there isn't, right? So if they are the same it should return false.

marco-c:
> Moreover we are adding to issues if testSizeDiff is returned true. But p1 && e1 < 0.8 this means that if p1 (eg. is aligned left) and e1 ((x1 - x2)/ (horizontal_width_child / 3)) is less than 0.8 (which means they are close in x direction) then we are adding to issues. While we should add to issues otherwise.

Why are they doing this then? Can you post a link to where testSizeDiff is used?

sagarvijaygupta:
> testSizeDiff should return true if there's a difference and false if there isn't, right? So if they are the same it should return false.

Yeah! I realised it afterwards! Thanks.

https://github.com/gatech/xpert/blob/63dcf7ffbb3cde33becad71d8c164f5e79c9d04d/src/edu/gatech/xpert/dom/layout/AGDiff.java#L63-L65

https://github.com/gatech/xpert/blob/63dcf7ffbb3cde33becad71d8c164f5e79c9d04d/src/edu/gatech/xpert/dom/layout/AGDiff.java#L226-L237

It might be I am missing something.

sagarvijaygupta:
@marco-c I get it. Sorry for the trouble. 
They are checking if one is said to be aligned and other is not, then they check how much aligned it is. If it is less than 0.8, that means that it is too much aligned and hence we have the issue since other one is not aligned. If it is not that aligned we consider it not that much aligned and skip the issue.

sagarvijaygupta:
Some findings while testing:
1. After the first part even if all the nodes were matched, in layout we match parents of nodes and siblings.
2. Siblings create higher number of issues than parents. It can be seen in the results gathered directly running the xpert itself also.

sagarvijaygupta:
@marco-c This PR is also ready!

PR #258, opened 2018-08-15
Port Crawler Improvements

### LINKED ISSUE #236: Port crawler improvements from coverage-crawler to autowebcompat crawler
https://github.com/mozilla/coverage-crawler/commit/f164a6de4a961277cb4006c6939290526bf5c955#diff-ee054b4b797f6c2a5743d2654764a8ed

https://github.com/mozilla/coverage-crawler/commit/7e4218a61d06da26477b6d740cd643f1808d29db?w=1

Shashi456:
Fixes #236

### DIFF
--- a/collect.py
+++ b/collect.py
@@ -10,12 +10,18 @@ import traceback
 from PIL import Image
 from lxml import etree
 from selenium import webdriver
+from selenium.common.exceptions import ElementNotInteractableException
+from selenium.common.exceptions import InvalidSelectorException
 from selenium.common.exceptions import NoAlertPresentException
 from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import NoSuchWindowException
+from selenium.common.exceptions import StaleElementReferenceException
 from selenium.common.exceptions import TimeoutException
+from selenium.common.exceptions import WebDriverException
 
 from autowebcompat import utils
+
+already_clicked_elems = set()
 
 MAX_THREADS = 5
 MAX_INTERACTION_DEPTH = 7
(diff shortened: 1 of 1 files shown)


collect.py
```diff
@@ -10,13 +10,19 @@
...
 from selenium.common.exceptions import NoAlertPresentException
 from selenium.common.exceptions import NoSuchElementException
 from selenium.common.exceptions import NoSuchWindowException
+from selenium.common.exceptions import StaleElementReferenceException
 from selenium.common.exceptions import TimeoutException
+from selenium.common.exceptions import WebDriverException
 
 from autowebcompat import utils
 
+already_clicked_elems = set()
```
marco-c: We have several crawler threads, so this can't be a global variable (otherwise the different threads will overwrite each other's variable).

collect.py
```diff
@@ -129,108 +135,124 @@ def was_visited(current_path, visited_paths, elem_properties):
...
+
+                    # We mark the current path as visited
+                    elems = get_elements_with_properties(driver, elem_properties, children)
+                    if len(elems) == 1:
+                        elem = child
+                        xpath = driver.execute_script(get_xpath_script, elem)
+                        break
+                    else:
+                        children_to_ignore.extend(elems)
+                else:
```
sagarvijaygupta: @marco-c This `else` is not aligned correctly with its corresponding `if`

Shashi456:
@sagarvijaygupta @marco-c  so i made ````already_clicked_elems```` a local variable and took care of indentation, there are redundancies in the exception blocks because we check if the xpath exists or not,  so there's a try-catch block for each conditional.

marco-c:
> there are redundancies in the exception blocks because we check if the xpath exists or not, so there's a try-catch block for each conditional.

What if you put the try before the if-else for xpath?

marco-c:
>  so i made already_clicked_elems a local variable

It can't be local to do_something, otherwise it gets cleared every time we click on something and thus it becomes ineffective.

Shashi456:
@marco-c where should i declare it instead? because one way i can think of is declaring it in the function which is calling it and pass it as a parameter

sagarvijaygupta:
@Shashi456 I think we should declare it in `run_test_both` and pass it to `jump_back` also.

marco-c:
> @Shashi456 I think we should declare it in run_test_both and pass it to jump_back also.

Yes, we should declare it in `run_test_both`.
Do we need it for jump_back? In jump_back we should already know exactly what we're going to click, right?

Issue #261, opened 2018-09-06
Throw a meaningful error in utils.read_labels when labels.csv is empty


poush:
Nothing serious but I believe that utils.read_labels should count for the case when labels.csv is empty and set an empty array/dict of labels instead of throwing error.

aru31:
I would like to work on this issue

marco-c:
Fixed by #269.

PR #266, opened 2018-12-05
Add handling for edge cases where file may not exist or be empty

aru31:
This is a fix for #261 accounting for blank files with `os.stat()` and preventing errors from being thrown.

### DIFF
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -188,13 +188,16 @@ def make_infinite(gen_func, elems):
 
 
 def read_labels(file_name='labels.csv'):
-    try:
-        with open(file_name, 'r', newline='') as f:
-            next(f)
-            reader = csv.reader(f)
-            labels = {row[0]: row[1] for row in reader}
-    except FileNotFoundError:
+    if os.stat(file_name).st_size == 0:
         labels = {}
+    else:
+        try:
+            with open(file_name, 'r', newline='') as f:
+                next(f)
+                reader = csv.reader(f)
+                labels = {row[0]: row[1] for row in reader}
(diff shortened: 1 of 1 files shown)


aru31:
@marco-c please review my PR

marco-c:
Instead of returning an empty dict, let's throw a meaningful error when the file is empty.

autowebcompat/utils.py
```diff
@@ -188,13 +188,16 @@ def make_infinite(gen_func, elems):
...
-    try:
-        with open(file_name, 'r', newline='') as f:
-            next(f)
-            reader = csv.reader(f)
-            labels = {row[0]: row[1] for row in reader}
-    except FileNotFoundError:
+    if os.stat(file_name).st_size == 0:
         labels = {}
+    else:
+        try:
```
marco-c: Don't catch the exception here, let it fail with FileNotFoundError if the labels file doesn't exist

autowebcompat/utils.py
```diff
@@ -188,13 +188,16 @@ def make_infinite(gen_func, elems):
...
 
 def read_labels(file_name='labels.csv'):
-    try:
-        with open(file_name, 'r', newline='') as f:
-            next(f)
-            reader = csv.reader(f)
-            labels = {row[0]: row[1] for row in reader}
-    except FileNotFoundError:
+    if os.stat(file_name).st_size == 0:
         labels = {}
```
marco-c: Throw an error here, saying that the labels file is empty.

marco-c:
Closing in favor of https://github.com/marco-c/autowebcompat/pull/267.

PR #267, opened 2018-12-06
Changes made for handling edge cases where file may not exist or be empty

aru31:
This is a fix for #266

### DIFF
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -188,13 +188,13 @@ def make_infinite(gen_func, elems):
 
 
 def read_labels(file_name='labels.csv'):
-    try:
+    if os.stat(file_name).st_size == 0:
+        raise ValueError('labels file is empty')
+    else:
         with open(file_name, 'r', newline='') as f:
             next(f)
             reader = csv.reader(f)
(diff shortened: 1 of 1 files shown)


autowebcompat/utils.py
```diff
@@ -188,13 +188,13 @@ def make_infinite(gen_func, elems):
 
 
 def read_labels(file_name='labels.csv'):
-    try:
+    if os.stat(file_name).st_size == 0:
+        raise ValueError('labels file is empty')
```
marco-c: Raise a normal Exception, as ValueError has a specific meaning (`Raised when an operation or function receives an argument that has the right type but an inappropriate value, and the situation is not described by a more precise exception such as IndexError.`).

aru31:
@marco-c i don't understand why coverage bot checks are failing. Please help

marco-c:
> @marco-c i don't understand why coverage bot checks are failing. Please help

It's because your change is apparently not covered by tests, but it's fine, those checks are not required.

marco-c:
Closing in favor of https://github.com/marco-c/autowebcompat/pull/269.

Please update the PR instead of opening a new one next time :)

PR #268, opened 2018-12-06
Load and run a trained model in the labeling UI

gabriel-v:
Add a `--eval_model` argument to `labels.py`. When this argument is
provided, the script will load that model and print its prediction for
each image pair displayed by the UI.

### DIFF
--- a/autowebcompat/utils.py
+++ b/autowebcompat/utils.py
@@ -80,6 +80,10 @@ def prepare_images():
 images = {}
 
 
+def load_pair(fname):
+    return [fname + '_firefox.png', fname + '_chrome.png']
+
+
 def load_image(fname, parent_dir='data_resized'):
     global images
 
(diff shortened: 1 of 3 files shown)


autowebcompat/utils.py
```diff
@@ -80,6 +80,10 @@ def prepare_images():
 images = {}
 
 
+def load_pair(fname):
```
marco-c: Since you're introducing this function, make use of it in train.py and pretrain.py too.

label.py
```diff
@@ -29,6 +32,32 @@
     random.shuffle(images_in_all_labels)
     images_to_show = images_not_in_all_labels + images_in_all_labels
 
+if args.model:
```
marco-c: It would be nice to move the evaluation code into a module, so that it can be shared by label.py and by whatever module we will implement in the future as a backend for the UI.

Issue #280, opened 2018-12-15
Out of memory error while training vgg16 and vgg19 with imagenet weights on Colab


sdv4:
As in issue #191, when training with input shape (224, 224) and the following command line arguments:

```!python3 train.py -n=vgg16 -o=sgd -ct='Y vs D + N' -bw='imagenet'```

I receive the following error messages, followed by the process being killed automatically:
```
Using TensorFlow backend.
tcmalloc: large alloc 6457049088 bytes == 0x4e70000 @  0x7f0ddb5c3001 0x7f0dd8bf8b85 0x7f0dd8c5bb43 0x7f0dd8c5da86 0x7f0dd8cf5868 0x5030d5 0x507641 0x504c28 0x502540 0x502f3d 0x506859 0x504c28 0x506393 0x634d52 0x634e0a 0x6385c8 0x63915a 0x4a6f10 0x7f0ddb1beb97 0x5afa0a
tcmalloc: large alloc 6457049088 bytes == 0x185ca6000 @  0x7f0ddb5c11e7 0x7f0dd8bf8a41 0x7f0dd8c5e7c0 0x7f0dd8c53ce5 0x7f0dd8cf64f3 0x5030d5 0x507641 0x504c28 0x502540 0x502f3d 0x506859 0x504c28 0x502540 0x502f3d 0x506859 0x504c28 0x502540 0x502f3d 0x506859 0x504c28 0x506393 0x634d52 0x634e0a 0x6385c8 0x63915a 0x4a6f10 0x7f0ddb1beb97 0x5afa0a
^C
```
It seems that another full dataset allocation is happening somewhere else, besides where it was fixed in #270.

PR #281, opened 2019-01-26
Fix for out of memory error issue #280

sdv4:
The problem was fixed by removing the creation of an array containing a vector for every image file in utils.py/get_ImageDataGenerator. The code was there in preparation for a call to fit() for the ImageGenerator  object. Upon inspecting the documentation for this method, it turns out that this call was not necessary for the image processing we are doing. 

I tested this on Colab for a few epochs on vgg16 using imagenet weights.


marco-c:
I wonder why this wasn't a problem in @sagarvijaygupta case (he was using Colab too).

> Upon inspecting the documentation for this method, it turns out that this call was not necessary for the image processing we are doing.

It isn't necessary for the image processing we are doing right now, but could be needed in the future if we try to do some. Anyway, we can bring it back if needed.

autowebcompat/utils.py
```diff
@@ -81,20 +81,12 @@ def prepare_images():
 
 
 def load_image(fname, parent_dir='data_resized'):
-    img = load_img(os.path.join(parent_dir, fname), target_size=(32, 24))
+    img = load_img(os.path.join(parent_dir, fname), target_size=(224, 224))
```
marco-c: Can you remove this change from the PR? It should be done in a separate PR, after this one.

label_persons/data
```diff
@@ -0,0 +1,3 @@
+Image Name,Label
```
marco-c: Can you remove this file from the PR?

sdv4:
I am going to open a new pull request that with a clean commit history and with the changes @marco-c specified above.

marco-c:
> Was he using Colab when the dataset was much smaller, before all of that additional labelling was done back in Oct 2018?

Yes, I think so.

> I am going to open a new pull request that with a clean commit history and with the changes @marco-c specified above.

It's better not to close a PR and open a new one in cases like this (well, basically in all cases :P). Otherwise you lose the history of discussion.

sdv4:
> It's better not to close a PR and open a new one in cases like this (well, basically in all cases :P). Otherwise you lose the history of discussion.

Oh ok, that makes sense. Thanks for letting me know; I will make sure to follow this procedure in the future!

PR #285, opened 2019-07-31
Docker and pipenv

gabriel-v:
This switches from pip to pipenv and adds a Dockerfile to the project root.

A `docker build` is also triggered on travis.


.dockerignore
```diff
@@ -0,0 +1,3 @@
+/data_resized
```
marco-c: You can take a few more things from https://github.com/marco-c/autowebcompat/blob/master/.gitignore (specific to this project) and https://github.com/mozilla/bugbug/blob/master/.dockerignore (generic stuff).

Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM python:3.7-buster
+
+RUN mkdir /autowebcompat
+WORKDIR /autowebcompat
+
+ADD Pipfile Pipfile.lock ./
+RUN pip install pipenv
+RUN pipenv install --system --deploy --ignore-pipfile
```
marco-c: Nit: you can install pipenv, run pipenv install, and then uninstall pipenv in the same line, so the resulting image will not have pipenv.

marco-c: Also use --no-cache-dir to avoid caching the package (not sure what's the equivalent for pipenv, if there is an equivalent).

You can also use `--disable-pip-version-check` to avoid useless logs.

marco-c (approved):

